PR #55, opened 2015-02-22
Lookup game in database based on the rom name, if the MD5 does not succeed

cplr:
This is necessary for ROMs that are not in the database. Right now this assumes USA for the region. While it is functional, there's probably a better way at getting the system than using TEMPsystemShortName in the SQL query.

Moving beyond this, it would be nice to show a popup that shows all the results based on the name, with their box art, and let the user pick the correctly matching game. Currently it just grabs the first result, like normal.


jasarien:
This isn't relevant anymore  since the whole import/lookup code has been rewritten. Games are looked up by title as well as hash now.

Issue #39, opened 2014-10-24
Controller support for iCade/8Bitdo NES30/Bluetooth keyboard


sojuicy:
Please consider adding controller support for the 8Bitdo NES30 controller, which can operate in icade or bluetooth keyboard modes.

Keloran:
👍

josejulio:
+1

josejulio:
is anyone working on this?

jasarien:
I don't have any plans to implement this myself. I don't have any of the listed controllers so wouldn't be able to test them. MFi game controllers are supported. If someone else wants to implement this and make a pull request I'll be happy to merge it in.

josejulio:
I just ordered a nes30 pro(which has an iCade mode), once it arrives i'll seriously consider to implement this myself. Thanks.

josejulio:
iCade mode works as a bluetooth keyboard, with events to handle the buttons pressed (https://github.com/stuartcarnie/iCade-iOS/blob/master/iCadeTest/iCade/iCadeReaderView.h)

Did some google searchs and dug a bit on the iOS sdk, and seems that detecting if a external keyboard is not possible (workarounds use UIKeyboardWillShowNotification, since the keyboard never shows i guess that it won't be the case), another way is to listed to some private API to know if an external keyboard is connected, but after a bit of though, an external keyboard doesn't mean that the user has an iCade controller. 

So i'm guessing that the user could actually choose if he wants to use an iCade controller (in settings) or have the iCade controller always listening.

Also, for the actual inner working, I'm thinking on create a classes that implements the GCControllerDirectionPad and GCControllerButtonInput to make it compatible with the actual implementation of PVControllerViewController.

May i know your thoughts on this?

josejulio:
I have raw implementation, needs to tune some details, and add a configuration for the controller, like Listen for iCade Controller, but it works, for anyone wanting to test it out, you can get it from https://github.com/josejulio/Provenance, i'll send a PR once I finish. Would be nice to hear comments about it.

axroberts:
Great work jose! Works like a charm with my SNES 8bitdo controller!

josejulio:
Thanks, though it still needs a bit more of work.

cortneytaylor:
I agree. It works great, with a few small issues you may already be aware of: 

1) The major one that i can see is that the controller doesn't disconnect from the game when you exit, which then makes it unavailable for the next game.

[...]

PR #78, opened 2015-07-13
Add support to iCade controller

josejulio:
-Setting to select if *listen* for iCade controllers and select what layout should be used (standard, 8bitdo).
-Option to *disconnect* the iCade.


josejulio:
Fixes #39

PR #103, opened 2015-09-20
Some tvOS controller fixes

kckd:
I haven't spent enough time to validate that these are correct in all cases, but in my testing on tvOS using a bluetooth controller, they work correctly and fix a couple issues.

One changes the way NES controller dpad events are processed, as zelda and other games were quite difficult to control, especially with the joystick.

The other change adds 1 to the player # for the SNES controller, as it appears to be 1 based. Without the change, the controller simply doesn't seem to work. 

It's entirely possible that these break the iOS onscreen controller, but at the very least they might be a starting point for fixing the issues I've noticed.


jasarien:
Thanks!

kckd:
Thank's for creating Provenance! Glad I could contribute.

PR #112, opened 2015-09-26
Diagonal controls from dpad and analog stick are not detected

jtrstz:
Diagonal controls (i.e. direction UP and Right at the same time) are being detected as either an horizontal or a vertical direction.

The following code is causing the issue by skipping pressed direction:
```
&& fabsf(xAxis) > fabsf(yAxis)
&& fabsf(xAxis) <= fabsf(yAxis)
```

### DIFF
--- a/Provenance/Controller/PVGBAControllerViewController.m
+++ b/Provenance/Controller/PVGBAControllerViewController.m
@@ -180,7 +180,7 @@
 
     float xAxis = [[dpad xAxis] value];
     float yAxis = [[dpad yAxis] value];
-    if (xAxis != 0 && fabsf(xAxis) > fabsf(yAxis))
+    if (xAxis != 0)
     {
         if (xAxis > 0)
         {
@@ -197,7 +197,7 @@
         [gbaCore releaseGBAButton:PVGBAButtonLeft forPlayer:player];
     }
     
-    if (yAxis != 0 && fabsf(xAxis) <= fabsf(yAxis))
+    if (yAxis != 0)
     {
         if (yAxis > 0)
         {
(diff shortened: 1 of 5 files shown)


jasarien:
Thanks, I thought something was up, but I couldn't pin-point it.

jasarien:
Hm, it still behaves kinda weird...

jtrstz:
In what situation? SNES emulator or another one?

jasarien:
I just pushed a commit adding some deadzones, give it a try, let me know what you think?

jtrstz:
I also did something similar but then I thought the value would need to be configurable because behaviour will depend on the controller brand...

jasarien:
It should be ok - all MFi controllers should be calibrated for physical deadzones anywya - the game controller framework normalises the values fro -1.0 to 1.0, so 0.5 should always be half the travel of the analog on all MFi controllers.

jtrstz:
I have something I didn't expect on my controller, the dpad is not digital! it's analog... So it's analog as a button can be... It feels a bit strange because you have to press it hard all the time to pass the 0.5.

jasarien:
Out of interest what controller is it?

jtrstz:
Mad Catz Micro C.T.R.L.i

jasarien:
I don't know what, if any, a 100% perfect solution would be.

jtrstz:
Fragmentation issue... well 0.5 should be fine with the majority. dpad should not be analog anyway.
Can add a way to configure the default value later.

PimCoumans:
I'm not sure if this is the right place to bring this up, but I'm experiencing some dpad deadzone issues. While navigating the menus on tvOS, everything works great, but in games I find I need to press harder after the initial physical feedback I've pressed hard enough. This is on the new Steelseries Nimbus by the way.

PR #129, opened 2015-10-05
Fixes iCade disconnecting when loading a state.

josejulio:


indirect:
I can confirm that this fixes the SNES30 iCade controller after loading states :tada:

PR #135, opened 2015-10-06
Fix build on tvos beta 3

tmm1:
Was getting lots of errors like this:

```
Target 'PVGBATV' has bitcode disabled (ENABLE_BITCODE = NO), but it is required for the 'appletvos' platform
```


jasarien:
I haven't had a chance to try with Beta 3 yet, is this a new constraint introduced to tvOS 3? What about the other emulator cores? Do they complain about bitcode? I don't know how bitcode affects emulation as I haven't had any chance to test it.

I'll leave this pull request open until I can investigate more.

tmm1:
I had to enable bitcode on all the targets, as you can see in the PR. Once I did this, the project built and I confirmed it works on the device.

This must be a new requirement in beta3 as I did not have any issues on beta2.

jasarien:
How does it affect emulation performance, if at all?

tmm1:
I did not notice a difference. Although, I'm not sure if a bitcode binary was transferred to the device or if that's something specific to the app store submission process.

maddox:
Yup, bitcode is required as of [tvOS beta 3](https://developer.apple.com/library/prerelease/tvos/releasenotes/General/RN-tvOSSDK-9.0/index.html). 

> Furthermore, bitcode is now required for all tvOS apps. All apps and frameworks in the app bundle need to include bitcode.

asos-jamesaddyman:
I've manually updated the project for bitcode support. Thanks for the heads up.

jasarien:
I've manually updated the project for bitcode support. Thanks for the heads up.

PR #146, opened 2015-10-16
More Settings UI fixes

saagarjha:
Made the iCade controller selection view controller pushed onto the navigation stack instead of presented modally, allowed for multiple chances to change the settings
Fixed a compile error for tvOS (oops…)


jasarien:
Could you explain to me again what benefit there is to pushing the view controller onto the stack vs a modal presentation? I'm not sure I understood from your original commit message.

Also, in future please clean up the pull requests so they don't contain changes to provisioning/codesigning in the project file.

If you can explain to me the benefits, tidy up the pull request and solve the merge conflicts, I'll take another look.

saagarjha:
I've replaced the .xcodeproj and .xworkspace files with the "official" ones, hopefully that resets the provisioning. In the future, though, is there an certain way you'd like me to reset the code signing?

As for the UI fixes, I changed selecting the iCade controller such that the user has the opportunity to change their decision before deciding to "save" it by navigating back, instead of immediately kicking the user out once they made the choice. I changed it to being pushed to the stack since it didn't really fit the modal presentation model. Since you've already fixed the tvOS compile bug, this push is completely up your preference between it being pushed or displayed modally.

jasarien:
Merge conflicts still haven't been resolved and the repo has moved on a bit since this pull request, so its being closed.

PR #148, opened 2015-10-16
ShortcutItemXXX API only for IOS9

elisaday:


jasarien:
Thanks for the heads up, fixed manually in latest commit in master.

PR #155, opened 2015-10-22
Fix for iCade 8bitdo Nes30 controller directions

StyloBill:
Directions fixed for the d-pad.
8Bitdo NES30 firmware v2.63

### DIFF
--- a/Provenance/Controller/iCade/iCadeReaderView.m
+++ b/Provenance/Controller/iCade/iCadeReaderView.m
@@ -22,8 +22,8 @@
 
 #import "iCadeReaderView.h"
 
-static const char *ON_STATES  = "wdxayhujikol";
-static const char *OFF_STATES = "eczqtrfnmpgv";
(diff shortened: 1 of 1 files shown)


jasarien:
My only question for this pull request is does this break NES30 compatibility for people with older firmware? I feel like this should be implemented in a way that the firmware version is checked and the appropriate key mappings are used. If I merge this  pull request as it is I'll just get a load of new issues from people with the older firmware complaining it doesn't work for them any longer.

StyloBill:
After more tests, it appears that the issue is not related to Provenance.
Actually it is an iOS problem : the controller-key mapping is dependant of the keyboard layout selected on iOS.
To correctly map the controller with the keyboard keys, the iOS keyboard layout must be on QWERTY.

A possible Provenance improvement might be to check the OS keyboard layout and to remap the keys accordingly.

I close the PR.

PR #158, opened 2015-10-28
Removed all deadzones as GCControllerAxisInput handles them out of the box

PimCoumans:
See https://developer.apple.com/library/prerelease/ios/documentation/GameController/Reference/GCControllerAxisInput_Ref/index.html#//apple_ref/occ/instp/GCControllerAxisInput/value for more info

### DIFF
--- a/Provenance/Controller/PVGBAControllerViewController.m
+++ b/Provenance/Controller/PVGBAControllerViewController.m
@@ -180,14 +180,14 @@
 
     float xAxis = [[dpad xAxis] value];
     float yAxis = [[dpad yAxis] value];
-    if (xAxis > 0.5 || xAxis < -0.5)
+    if (xAxis != 0)
     {
-        if (xAxis > 0.5)
+        if (xAxis > 0)
         {
             [gbaCore pushGBAButton:PVGBAButtonRight forPlayer:player];
             [gbaCore releaseGBAButton:PVGBAButtonLeft forPlayer:player];
         }
-        else if (xAxis < -0.5)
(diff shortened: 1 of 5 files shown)


jasarien:
This appears to be a hardware fragmentation issue. While the framework does indeed handle deadzones, as you say, the reason these 'artificial' deadzones were added was to make the analogue directional inputs less sensitive. Even with the framework-handled deadzones, moving the analogue controller all the way to the left and slightly down (eg, left: 1.0 and down: 0.1), causes it to register as a down-left input which is not intended by the user. Without the 0.5 deadzones, control is unpredictable and often frustrating, which is why they were added in the first place.

As for the hardware fragmentation, it's down to each manufacturer to decide how much pressure is necessary for a control to move from 0.0 to 1.0 on the normalised scale. Some controllers will require more pressure than others. We could drop the value which might make it easier for some controllers, but then maybe other controllers act unpredictable again. 0.5, being half way between 0 and 1 seemed like the best compromise at the time.

Perhaps the best option will be to make this a user-configurable setting.

jasarien:
I'm going to close the pull request because it just reverts to code that was in place before the 0.5 deadzones, and many people complained about poor control experiences without the 0.5 deadzones.

PimCoumans:
Did not know this was the reasoning behind the deadzone of 0.5, felt like an arbitrary magic number to me. Making it user configurable would handle this case better. Needing to push harder on the dpad than the actual physical feedback feels weird on the Steelseries Nimbus. Thanks for the fast and elaborate reply!

PR #171, opened 2015-11-03
Reduce startup time for large libraries. 

anwnn:
… on every iteration.

Greatly reduces startup time for large libraries.


anwnn:
Hey James,

Would you mind amending the author on that commit? I accidentally committed
with the work email address, rather than my github one ;)

I'd appreciate it if you could change it to "James G. <james@veldt.com>".

Thanks for your work on Provenance. You do have some fans here at Apple,
and more than one contributor. =)

Cheers,
--James

PR #201, opened 2015-11-10
Exposed Sega SG-1000 support

apexad:
https://en.wikipedia.org/wiki/SG-1000
Genesis Plus GX has support for it, so why not expose it. I've tested this and it works.

The only thing I don't quite understand is what PVDatabaseID should be set to in systems.plist for this system.  It is currently 33, the same as Sega Genesis. Let me know what it should be and I can update it to the correct value.

### DIFF
--- a/Provenance/Emulator/PVEmulatorConfiguration.h
+++ b/Provenance/Emulator/PVEmulatorConfiguration.h
@@ -43,6 +43,7 @@ extern NSString * const PVGBSystemIdentifier;
 extern NSString * const PVGBCSystemIdentifier;
 extern NSString * const PVNESSystemIdentifier;
 extern NSString * const PVFDSSystemIdentifier;
+extern NSString * const PVSG1000SystemIdentifier;
 
 @interface PVEmulatorConfiguration : NSObject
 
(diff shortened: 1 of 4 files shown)


apexad:
Sorry for all the title changes. Since this is a bit of an unknown system that released at the same time as Famicom in Japan, I wanted to tag it for anyone who might not have heard of it.  I decided against it though and figured anyone could do the research if they so desired.

I also added commits to update the README and did end up making PVDatabaseID 34 instead of 33 so it is different.

jasarien:
35 is the correct ID for the SG-1000. You get the ID from the OpenVGDB.sqlite file - open it in sqlite browser like Base.app or on query it with sqlite on the command line. There is a `SYSTEMS` table with the necessary info.

apexad:
Did you want me to change this? (To 35) or are you?

jasarien:
I've done it, thanks for the pull request :)

apexad:
Is that change committed?

jasarien:
It is now. Not sure what happened there, I definitely committed it. Must have gotten lost somehow :/

PR #210, opened 2015-11-14
Add ability for two players to use one controller

ericyanush:
Change game controller setup to allow two players to share one controller.
Game controller selection settings already allows this, however when a single controller is selected for both players, input is only used for player 1. This change allows both players to share a single controller.

### DIFF
--- a/Provenance/Controller/PVControllerViewController.m
+++ b/Provenance/Controller/PVControllerViewController.m
@@ -22,6 +22,8 @@
 @property (nonatomic, strong) NSArray *controlLayout;
 @property (nonatomic, assign) BOOL touchControlsSetup;
 
+@property (nonatomic) BOOL bothPlayersUseSameController;
+
 - (void) listenForICadeControllers;
 
 @end
@@ -35,6 +37,7 @@
 		self.controlLayout = controlLayout;
         self.systemIdentifier = systemIdentifier;
         self.touchControlsSetup = NO;
+        self.bothPlayersUseSameController = NO;
 	}
 	
 	return self;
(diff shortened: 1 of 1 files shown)


jasarien:
What's the use-case for this, exactly? It's not like two people can hold the controller at the same time?

ericyanush:
I created this patch so I could play two player nes games (super mario, donkey kong, dig dug, etc) with only a single controller. Because these games only have one active player at a time, players can pass off the lone controller at the end of turn, requiring only one controller for both people to play.

keithwahn:
This seems like a decent idea, considering the controllers are like $50 a pop.

jasarien:
This patch isn't necessary. You can just set the same controller to player 1 and player 2 in the controller settings  like so: same controller and get the same benefit.

ericyanush:
The patch is necessary. Although you can select the same controller for both players, it doesn't work. (Tested with a Steelseries nimbus). The reason it doesn't work is because of the way the controller setup happens.  On line 380 of PVControllerViewController, in setupGameController, when the method tries to determine which player to attach the button press/release handlers to, it will always attach to only player one because the check for `if (controller == [[PVControllerManager sharedManager] player1] || controller == self.iCadeController)` will always be true, so the else case for player 2 will never execute, and the handlers will never attach for player 2, resulting in inputs for player 2 always being ignored.

jasarien:
They do in the new code. I just committed a big rewrite of the controller handling, for a separate issue, So the code here wouldn't be able to merged in without significant changes anyway. I was in the middle of the rewrite when you submitted the pull request.

Anyway, I tested with the Siri remote as the only connected controller and I was able to play 2 Player Super Maro Bros on the NES with the single Siri Remote alone. I also tested with a Steelseries Startus XL, which is very similar to the Nimbus, and it also worked for player 1 and 2 when it was set for both players.

keithwahn:
:+1: I just found this project and can't wait to try it out!

ericyanush:
@jasarien I apologize, I hadn't seen your latest commits. I tested them, and they are also working with the nimbus. Thanks for all your hard work on this project, it's awesome to have these emulators running on apple tv!

jasarien:
No need to apologise. I only committed them a few minutes before commenting here.

PR #218, opened 2015-11-20
Add controller mapping for SteelSeries Free

SFrost007:
This change adds a more appropriate controller mapping for the SteelSeries Free controller - the code is based on the iCade8Bitdo implementation. I set the buttons up to mirror the layout of the SNES pad as this most closely matches the physical layout of the controller, but tested with the Genesis emulator and it also works well.


appletv4:
Can you also fix this for the nimbus controller 
NES button layout is all messed up

SFrost007:
Unfortunately I don't have a Nimbus controller to test with, or an Apple TV. Have you tried this pull request? There's a chance it may work if they've used a similar button mapping for the two controllers.

jasarien:
@appletv4 
If I understand correctly, the Steelseries Free is not an MFi controller - it existed before the MFi controller program was started and before the GameController Framework was released. This pull request allows it to work by wrapping it with a GameController compatible interface, just like the iCade controllers.

The Steelseries Nimbus is an MFi controller and follows the specifcation of all other MFi controllers - ie, the same button layout, the same button mapping and the same common API. The code in this pull request will have no effect on the Nimbus or any other MFi controller.

z3r01:
question , I'm not a programmer but love to test things and this is great! 
the support for other controllers like playstation controllers , thats done by modifying the bluetooth which can't be done unless the ate is jailbroken right?

jasarien:
Correct.

z3r01:
Hey jasarien, awesome work here as I use your emulator on both iPad pro and the new atv4. Just wanted to add (although I'm no coder here , and I would like to contribute by testing) I found that when connecting the steelseries nimbus controller , there was a delay on the buttons when pressed

PR #231, opened 2015-11-29
Use Nintendo’s 4-button to 2-button mappings

jonathantneal:
Nintendo recommends resting the thumb over the west and south buttons on a 4-button SNES-type controller in place of B and A on a 2-button NES-type controller. This helps prevent [repetitive strain injury and wrist fatigue](https://en.wikipedia.org/wiki/Nintendo_thumb) by keeping the thumbs in an inward, upright position.

snes-to-nes


appletv4:
Hi @jonathantneal can you do this for gameboy and gameboy advance , thanks

PR #239, opened 2015-12-12
Fix Real Time Clock (RTC) by always setting it to true. Fixes #151

### LINKED ISSUE #151: GBA emulator - Real Time Clock (RTC) support
Its not high priority since only a few GBA games use Real Time Clock (RTC) but it seems like the GBA emulator doesn't support it. It would be great to have that for Pokemon hack games that support it (like Pokemon Liquid Crystal) or Boktai games (there are some IPS patches for those ones, though).

Thanks!

apexad:


jasarien:
Does overriding this to always be true have any negative effects? I feel a little reluctant to just always return true, it seems like it's settable for a reason?

apexad:
Does not seem like it has any negative effects. I'm still able to play games that don't use the Real Time Clock just fine.

PR #252, opened 2015-12-23
Allow for Opacity 0 for Controller Overlay

rnaud:
When using a MFI Controller It is nice to be able to fully remove the controller overlay.


jasarien:
When an MFi controller is connected the on screen controls are hidden, so no need to set their opacity to zero.

rnaud:
Haha, should've tried before making some commits ;) 
Closing.

PR #254, opened 2015-12-24
update Realm frameworks to version 0.97

jpsim:
Realm now officially supports AppleTV as of 0.97 ([source](https://realm.io/news/realm-objc-swift-0.97.0/)) and there have been lots of bug fixes and performance improvements since the last version used in Provenance (0.95.1-ish).

To keep things simple, I've opted to keep the db schema exactly as it was (no nullable properties) to avoid having to perform a migration. But if you want to make use of nullable properties in the future, now you can. Just keep in mind that you'll need to bump the schema version if and when you choose to do so.

Nice project by the way! We're having a blast playing NES games over the holidays :smile:.


apexad:
I assume @jasarien is looking at this as well, but one glaring thing I see is that you've committed Realm-iOS/Realm.framework/CHANGELOG.md

While it's great to know the changes in version 0.97, I don't think this file is should be included (and it was not there before).

jpsim:
@apexad this file is bundled as part of the official Realm binary framework releases. It's safe to remove if you'd rather not have it in there, but I think a more reasonable approach would be to not check in the binary, but rather use a dependency manager like CocoaPods or Carthage to incorporate it into the project.

apexad:
@jasarien has said in the past he does not want to use CocoaPods or Carthage (or anything that requires setup outside of Xcode), and I tend to agree with him

There are a couple other files as well:

Realm-iOS/Realm.framework/LICENSE
Realm-tvOS/Realm.framework/CHANGELOG.md

jpsim:
> @jasarien has said in the past he does not want to use CocoaPods or Carthage (or anything that requires setup outside of Xcode), and I tend to agree with him

You could also add the realm-cocoa repo as a submodule and build from source to avoid this:

image

But that's up to you.

> There are a couple other files as well:
> 
> Realm-iOS/Realm.framework/LICENSE
> Realm-tvOS/Realm.framework/CHANGELOG.md

Yes, these are safe to delete.

apexad:
Because of the way git works, if the intent is to not have those 3 files in the main repository, you need to delete them in your branch. When I say delete, I mean force a new commit so they are not uploaded at all.

jpsim:
Ok, this has gotten out of hand... now you feel the need to explain to me how git works.

If one of the project maintainers concur that the `CHANGELOG.md` files should be nuked from this commit, I'm happy to do so. The license files should stay in there as Realm should be distributed along with the Apache 2.0 license.

@jasarien is there anything I can do to move this along? I strongly encourage you to update Realm to 0.97 for official tvOS support. Whether you do that through this PR or otherwise is of no importance to me.

jasarien:
Sorry for the delays guys. I'm really busy recently. I'm moving jobs and homes so my attentions are elsewhere for the time being. Sit tight, things will settle down soon.

jpsim:
No worries, thanks for chiming in @jasarien and good luck with everything!

jasarien:
Updated Realm to 0.98.1 manually.

jpsim:
Excellent!

PR #298, opened 2016-03-08
Adjustments to control layout

gabrieloc:
## What this does
Makes minor adjustments to how elements are laid out on screen, making it more comfortable to play in portrait orientation.

group2

group
iPad, iPhone 6+, and iPhone 4s in portrait and landscape

### DIFF
--- a/Provenance/Controller/PVControllerViewController.m
+++ b/Provenance/Controller/PVControllerViewController.m
@@ -119,13 +119,18 @@
         {
             NSString *controlType = [control objectForKey:PVControlTypeKey];
             
+            BOOL compactVertical = self.traitCollection.verticalSizeClass == UIUserInterfaceSizeClassCompact;
+            CGFloat kDPadTopMargin = 16.0f;
+            CGFloat controlOriginY = compactVertical ? kDPadTopMargin : CGRectGetWidth(self.view.frame) + kDPadTopMargin;
+			
             if ([controlType isEqualToString:PVDPad])
             {
                 CGFloat xPadding = 5;
-                CGFloat yPadding = 5;
+                CGFloat bottomPadding = 16;
                 CGSize size = CGSizeFromString([control objectForKey:PVControlSizeKey]);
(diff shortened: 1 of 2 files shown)


pmp174:
Does this work on all cores or only the gameboy one?

gabrieloc:
Tested on Gameboy, GBC, GBA

Assuming it works fine on all others, since it looks like controls are all laid out in the same place

oliviercharavel:
That's pretty good! 😊 I've noticed a little bug when rotating the device during gameplay will move the buttons closer to the top of the screen every time though.

gabrieloc:
@oliviercharavel Thanks! I probably should've documented the landscape changes. The thinking was similar to the portrait changes:

group

apexad:
Just a thought, you may want to squash the 'indentation' commits. Otherwise this looks really good.

oliviercharavel:
@gabrieloc So here is more about what I was talking about :smile: .

If I do the following:
- Launch Provenance in portrait.
- Launch a game.
- Rotate the device to landscape while in game.
- Quit the game by going back to the menu.
- Stay in landscape and launch the game again or another one (notice how the buttons moved closer to the top):
 
screenshot 2016 03 10 08 37 21

And if I rotate back to portrait while in game:
provenance-issue

This is how portrait should look like:
provenance-normal

To avoid the issue: When quitting a game and you're back  to Provenance library, rotate the device to portrait and launch the game again or another one. Tested on an iPhone 6s Plus.

gabrieloc:
@apexad Squashed those useless commits.

@oliviercharavel It looks like we create and layout controls in the same place (`-setupTouchControls`), but rely on an ivar for deciding wether or not to skip the implementation of `-setupTouchControls`, despite there being a conditional in place allowing for views to be laid out without being recreated. This was preventing views from being laid out a second time, eg. after we change orientation.

It looks like the ivar was only being used in this one place, so I've removed it and de-indented the entire block that used it. Let me know what you think.

PR #301, opened 2016-03-12
Increase buffer size to fix controller issue

kyuzumaki:
Fix for issue where emulator stalls on hitting buttons quickly. Buffer size increased to prevent it clearing too often when buttons are pressed very quickly.

### DIFF
--- a/Provenance/Controller/iCade/iCadeReaderView.m
+++ b/Provenance/Controller/iCade/iCadeReaderView.m
@@ -125,7 +125,7 @@ static const char *OFF_STATES = "eczqtrfnmpgv";
     }
     
     static int cycleResponder = 0;
-    if (++cycleResponder > 20) {
+    if (++cycleResponder > 100) {
         // necessary to clear a buffer that accumulates internally
         cycleResponder = 0;
         [self resignFirstResponder];


apexad:
This looks like a very simple change. Any reason this has not been merged yet?

jasarien:
Because I haven't tested it due to the fact I don't have iCade hardware and don't know much about iCade in general. Also, the change is in 3rd party code, specially the open source iCade library, which is probably out of date itself. The library should be updated first and then the change reapplied if necessary. Finally, some justification about why the change is necessary, if the original iCade library owner didn't include it.

kyuzumaki:
This fix only applies to version 1.2.5

iCade:
iCade is just a clever use of keyboard inputs. One key denotes button down another denotes button up e.g. left is "A" & "Q". The class reads the input string and determines which keys are currently up and down based on the input string.

Reason:
When the iCade library flushes the input buffer it causes the emulator to jam. I just made it bigger to stop this happening constantly during gameplay. Its literally a change of one number. I have an 8bitdo iCade controller and can confirm this works at least for the devices I have access to (iphone 5, ipad2 and iphone 6 plus).

Original Dev:
I have no idea why the original developer chose 20 as the buffer size it seems to be an arbitrary choice. It does not account for the high rate of input when playing street fighter or r-type. He has not updated this library in 5 years https://github.com/stuartcarnie/iCade-iOS

jasarien:
Why does it only apply to 1.2.5?

kyuzumaki:
Just because iCade does not work at all in 1.2.6. The code will still be fine but its never actually called. I think when the input method was changed to reduce lag the iCade part stopped working.

jasarien:
Yep you're right about that.

PR #314, opened 2016-03-24
Added new iPad Pro 83.5px@2x app icon

capnslipp:
Added app icon for iPad Pro (83.5px@2x = 167px), now required when uploading ad hoc builds to iTunes Connect.

Recreated the icon using the `bg.png` resource from the `TVAssets.xcassets`, combined with the `pad.png` resource.  Double-checking with PS's Difference blending mode against the `Icon-60@3x.png` icon scaled-down to 167x167, all lines are within a sub-pixel, and all color values are identical or ±1/256 channel value.


capnslipp:
Cheers!

PR #328, opened 2016-03-31
Added support for Gameboy 128k save files

jkjwong:
Some Gameboy games such as Pokemon and Little Sound DJ (http://www.littlesounddj.com/) require 128k flash save files. Provenance only supports 32k currently.

Libgambatte core has already been updated to support 128k saves but a lot of emulators (not just Provenance) have not yet updated the libraries with these changes.

I have added the changes needed; tested with LSDJ and it works great.

### DIFF
--- a/PVGB/GB/libgambatte/src/mem/cartridge.cpp
+++ b/PVGB/GB/libgambatte/src/mem/cartridge.cpp
@@ -609,7 +609,28 @@ LoadRes Cartridge::loadROM(std::string const &romfile,
 		default: return -1;
 		}*/
 
-		rambanks = numRambanksFromH14x(header[0x147], header[0x149]);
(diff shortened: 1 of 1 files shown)


clobber:
@jkjwong I don't see this code upstream at https://github.com/sinamas/gambatte

jkjwong:
@clobber it can be found here - https://github.com/sinamas/gambatte/blob/master/libgambatte/src/mem/cartridge.cpp

clobber:
@jkjwong No it isn't, and there is code that is clearly commented out for what I'm sure is a good reason.

apexad:
This code is actually from this file:
https://github.com/libretro/gambatte-libretro/blob/master/libgambatte/src/mem/cartridge.cpp

PR #332, opened 2016-04-03
Add better support for axis directions on extended game controllers

thedrick:
This adds improved logic for controllers with thumbsticks to determine the current direction of either stick. Before this change, the logic alwasy preferred the yAxis of the thumbstick making it difficult to move along the xAxis in games. I started by adding the directional support to the SNES emulator since that's what I've been playing, but I put the logic into a shared class so it can be easily used in the other emulator cores. Please let me know if I put files in the right places / if I should add this axis support to the other emulator cores as well. Thanks!

### DIFF
--- a/PVSNES/PVSNES/SNES/PVSNESEmulatorCore.mm
+++ b/PVSNES/PVSNES/SNES/PVSNESEmulatorCore.mm
@@ -30,6 +30,7 @@
 #import "OETimingUtils.h"
 #import <OpenGLES/EAGL.h>
 #import <OpenGLES/ES3/gl.h>
+#import "PVGameControllerUtilities.h"
 
 #include "memmap.h"
 #include "pixform.h"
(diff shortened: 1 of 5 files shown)


svanimpe:
@thedrick This issue was the first thing I noticed when trying Provenance for Apple TV. It's a great app, I'm happy to have found it, but not all games are playable due to this issue. For example, I found Street Fighter II for SNES to be impossible to play properly. I'm using a Nimbus controller and most of the time, the character either jumps or crouches, instead of moving forward (when using the analog stick).

I've also had issues where a game (Sunset Riders) requires Up as an input and I had trouble inputting it using the analog stick, as it had to be exactly up in order for it to be recognized.

I'm really hoping there's a way to resolve these :)

svanimpe:
@thedrick This issue was about the first thing I noticed when trying out Provenance for the first time. Most SNES games are unplayable with the analog stick as a result of this. I tried out your code and while it does fix some issues, it also creates a new one: with your changes, it's impossible to do diagonal movement with the analog stick. I'm not sure how the SNES handles diagonal movement, but if is possible to do it with the D-pad, so the stick should support it too. I'm guessing you'll need to divide the circle into 8 (instead of 4) sectors, use a bit of trigonometry to find out what sector the stick is in, and report diagonal movement so it can trigger both a left or right and a up or down button.

jasarien:
@thedrick if you'd like to merge @svanimpe pull request on your fork so that it shows up in the changes on this PR, then I can check it out.

thedrick:
@jasarien, @svanimpe thanks, sorry about the delay. I've just merged the PR, so this should be updated now :)

jasarien:
Works nicely, cheers guys.

PR #333, opened 2016-04-04
Use default animation when focusing in the game library

thedrick:
This changes the game library cell to use the built-in focus engine / animation. I added a category on UIImage to generate an image with a color and text so focus would apply to games without artwork as well. I also wanted to replicate the same logic that the tvOS home screen uses for the titles, so they don't display by default, and instead animate it when that item comes into focus.

This is just a proposal, the code is a little messy and I'm happy to clean it up if this is something you're interested in doing with the project. Here are screenshots of tvOS and iOS versions

simulator screen shot apr 3 2016 7 44 28 pm

simulator screen shot apr 3 2016 7 52 03 pm


jasarien:
Did you manage to solve the aspect ratio issue with the built in focus effect? That's the main reason I implemented a custom version. Example: Genesis artwork generally is a tall rectangle shape, while SNES are long rectangles, GBA is square, etc. The issue I saw in the past was that the built in focus effect would ignore the `contentMode` property on the image view and would make everything square, which ended up cropping the tops and bottoms off the genesis images and the sides off the SNES images.

thedrick:
@jasarien yeah, instead of relying on the contentMode to take care of resizing the image, I manually resize the imageView to fit while maintaining the aspect ratio of the original image. I also made the missing artwork view an image itself so the focus engine could do the built in transition

jasarien:
Awesome, I'll give a try.

jasarien:
I love it.

thedrick:
YAY! thanks :D

PR #336, opened 2016-04-07
Fix #262: Hide unnecessary touch controls if system is Game Boy or GBA

### LINKED ISSUE #262: unnecessary touch controls when using mfi controller
When using an mfi controller like the Logitech Powershell for the iPhone 5 on Provenance v1.2.6, the soft buttons "select" and "start" hover the gameplay screen, even if they are unnecessary because they are mapped to the X and Y hardware buttons.

img_0285

apexad:
Tested and working. Fixes #262.

### DIFF
--- a/Provenance/Controller/PVControllerViewController.m
+++ b/Provenance/Controller/PVControllerViewController.m
@@ -388,7 +388,9 @@ void AudioServicesPlaySystemSoundWithVibration(int, id, NSDictionary *);
     [self.leftShoulderButton setHidden:YES];
     [self.rightShoulderButton setHidden:YES];
     
-    if ([controller extendedGamepad])
+    //Game Boy, Game Color, and Game Boy Advance can map Start and Select on a Standard Gamepad, so it's safe to hide them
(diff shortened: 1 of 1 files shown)


Provenance/Controller/PVControllerViewController.m
```diff
@@ -388,7 +388,8 @@ - (void)hideTouchControlsForController:(GCController *)controller
     [self.leftShoulderButton setHidden:YES];
     [self.rightShoulderButton setHidden:YES];
     
-    if ([controller extendedGamepad])
+    //Game Boy, Game Color, and Game Boy Advance can map Start and Select, so it's safe to hide them
+    if ([controller extendedGamepad] || [self.systemIdentifier isEqualToString:@"com.provenance.gba"] || [self.systemIdentifier isEqualToString:@"com.provenance.gb"] || [self.systemIdentifier isEqualToString:@"com.provenance.gbc"])
```
thedrick: suggestion: these values are defined in `PVEmulatorConstants` so using the variable name would help future proof this. Also might be easier to just add these into an array and call `[array containsObject:self.systemIdentifier]`

apexad:
Good call, fixed.

PR #349, opened 2016-05-13
Fix: Crash when opening Game Speed menu on iPad

werminghoff:
Added UIPopoverPresentationController sourceRect / sourceView parameters which are required on iPad when using UIAlertController's UIAlertControllerStyleActionSheet style.


jasarien:
Thanks!

PR #355, opened 2016-05-30
Fix MFI controller lag on SNES games

ubnt-tonytoole:
Fix MFI controller lag on SNES games

### DIFF
--- a/PVSNES/PVSNES/SNES/PVSNESEmulatorCore.mm
+++ b/PVSNES/PVSNES/SNES/PVSNESEmulatorCore.mm
@@ -131,11 +131,11 @@ NSString *SNESEmulatorKeys[] = { @"Up", @"Down", @"Left", @"Right", @"A", @"B", 
 - (void)executeFrame
 {
     IPPU.RenderThisFrame = YES;
-    S9xMainLoop();
     if (self.controller1 || self.controller2)
     {
         [self pollControllers];
     }
+    S9xMainLoop();
 }
 
 - (BOOL)loadFileAtPath:(NSString *)path


ubnt-tonytoole:
This is a stupid simple update, but it does fix the issue 100% for SNES games - poll the controller just before the SNES logic loop, and not after which makes the controls lag behind the game by a frame or more.

ubnt-tonytoole:
^^ github on mobile has some issues it seems.

lyret:
I implemented and tested this on my fork today, It works wonders.

ubnt-tonytoole:
I am pleased.  There is nothing more frustrating than laggy controls - it is one of my pet peeves and prevented me from using the emulator when I first installed it.  When I decided to take a look, it took what - 3 minutes to find and fix the problem.  Not sure why it was never fixed as I saw many lengthy threads regarding it in the issue tracker.

My family and I do not play anything other than SNES, so I never tested other cores but after taking a quick look, it appears NES could also benefit from this style of update as well.

Kudos to Jasarien for the awesome emulator and hard work put into it!

kristjen-kjems:
Would love to see this pushed to master! Thanks ubnt!

jasarien:
The [SNES9x porting guide](http://src.gnu-darwin.org/ports/emulators/snes9x/work/snes9x-1.51-src/doc/porting.html) states that controller input should be polled *after* running the `S9xMainLoop()` function. I assume this is because polling before the frame is executed is like trying to preempt the future.

> In your main emulation loop, after S9xMainLoop function is called, check your system's keyboard/joypad, and call S9xReportButton function to report the states of the SNES joypad buttons to Snes9x.
> 
> ```
> void MyMainLoop (void)
> {
>     S9xMainLoop();
>     MyReportButttons();
> }
> ```

Provenance used to do event-based controller handling meaning button presses were reported to the emulator as they happened rather than the emulator polling the controller. I think the latter is more performant, but it didn't appear to have made much of a difference to the latency issue people are perceiving.

I'm reluctant to change the order of controller polling so that it goes against the porting guide.

ubnt-tonytoole:
I believe the porting guide has errors - if you look at the Windows code [here](http://src.gnu-darwin.org/ports/emulators/snes9x/work/snes9x-1.51-src/win32/wsnes9x.cpp) you will see the controller is indeed polled right before S9xMainLoop(), contradicting the porting guide.

jasarien:
Hm... Appears to be the same in the MacOSX port too. Can't argue with that evidence.

ubnt-tonytoole:
There does appear to be an exception - the msdos port polls controllers after S9xMainLoop(), however this appears to be due to debug, and UI support, plus the loop is very tight, not creating any perceivable latency between the controller and S9xMainLoop calls.

fengb:
I've never heard of a game loop putting input logic *after* game logic. Sources:
* https://en.wikipedia.org/wiki/Game_programming#Game_structure
* http://gameprogrammingpatterns.com/game-loop.html
* http://gamedev.stackexchange.com/a/656

jasarien:
Yet more irrefutable logic.

jasarien:
I did the change for NES too. It appears that all the other systems (GenesisPlus, Gambatte and VBA) provide polling callbacks which I assume are called in the best places for their respective systems.

maddox:
I'm on `master` as of today, and I'm still seeing a ton of lag while playing Super Mario World. I've just updated my Nimbus controller as well. Did this pull solve all of the lag, or was it even more laggier than what I see right now?

Playing SMW with the lag I'm seeing is not very possible though. Did I maybe miss something?

ubnt-tonytoole:
Double check your branch - what do you have for the `executeFrame` function in `PVSNES/PVSNES/SNES/PVSNESEmulatorCore.mm` ?

appletv4:
@maddox you know more about Xcode than me (love your channels app) but are you sure your using the release version and not the debug version , it should tell you in provenance app settings

For me after the new provenance update its feels like I'm playing on the ps4 , lag free

I also have the nimbus controller 

It should say in settings 

Version : 1.2.7(411)
Mode : RELEASE

niltz:
I tried this and I was getting button release lag (I would release the button, and it would take a while for the emulator to respond to it). So I tried putting the polling both before AND after the main loop, and then I started to get sporadic behaviour. Does anyone else get the "button release lag" now?

maddox:
Dang! I swore that I was on up to date `master` and that I was using `release`.

But I just did everything all over again from scratch on my dev Apple TV (I originally did this all on my bedroom Apple TV), and it works great now. Also, I noticed that kicking off the web server is in the Settings now. This wasn't the case when I was doing it on my bedroom TV. It was in the old spot on the main tab.

So I clearly was doing something totally wrong.

Thanks everyone.

PS: anyone using 8bitdo iCade pads? How's the latency/lag?

appletv4:
The server thing In settings was added June 10 a day after my comment above or on the same day ,
The lag fix was added June 9 ,  but if you don't notice lag now perhaps you were on debug 

Icade only works on iOS for now not  Apple TV , but in the TVOS BETA 10 announcement I saw a new feature called " Bluetooth accessories" I wonder if that's support for icade ?  Or not MFI  controllers 


Seems like the people who use icade controllers and know how to code don't have an Apple TV , 
I have an snes30 8bitdo controller so I've been waiting since the Apple TV came out

Issue #173, opened 2015-11-03
Rename games in AppleTV UI


cbjornn:

jasarien:
This already a feature. Press and hold a game. 
![image](https://cloud.githubusercontent.com/assets/104444/10917041/a18e0a74-8256-11e5-9eac-5abe8c5c13a1.png)

cbjornn:
Is this feature available for the Apple TV version though?

asos-jamesaddyman:
No, but you didn't specify Apple TV in your issue report.

It's not available on Apple TV for a few reasons:

- There is no Asset Library on tvOS, meaning no way to choose a picture for cover art
- The text input on tvOS is diabolical, and there's no Siri Dictation API or bluetooth keyboard support, so I left it out.

jasarien:
No, but you didn't specify Apple TV in your issue report.

It's not available on Apple TV for a few reasons:

There is no Asset Library on tvOS, meaning no way to choose a picture for cover art
The text input on tvOS is diabolical, and there's no Siri Dictation API or bluetooth keyboard support, so I left it out.

chriscthompson:
Diabolical is the perfect word to describe the destruction they did to text input on tvOS. Lol

kudit:
I wonder why they didn't keep the grid layout of characters as it would be even easier to navigate with the new siri remote.

chriscthompson:
Or let's us do text to speech like iPhone text input

Or update the remote app to allow for easy text entry

Or allow iCloud sync of keychain passwords

Or allow Bluetooth keyboards

There are a lot of options for   to make this easier for everyone...

Hanterdro:
"The text input on tvOS is diabolical, and there's no Siri Dictation API or bluetooth keyboard support, so I left it out."

There is since iOS 9.2 an remote app with an very comfortable software keyboard.

ajmi982:
now that tvos 9.2 supports BT keyboards, could this be reopened?  I have a number of titles that dont have artwork, would be great to rename the titles and add artwork.

Or perhaps we can do it from the Import Roms screen in our browser? I was trying to find where the images are stored but cant find them

apexad:
This issue can now be closed since it is now possible to rename games on tvOS.

ajmi982:
Hi looks like lots of progress. He can I get latest build

jasarien:
[Building Provenance](https://github.com/jasarien/Provenance/wiki/Building-Provenance)

PR #318, opened 2016-03-28
Recent games

leolobato:
Recently played games section on library.

See #317

PR #358, opened 2016-06-09
Rename games on tvOS

leolobato:
Present an alert controller with a text field
Return an array of index paths when searching for a game on the data source
(supports game being on multiple sections, e.g. 'recently played')

Will resolve the part of renaming on #173, still need a way to pick images.

### DIFF
--- a/Provenance/Game Library/PVGameLibraryViewController.m
+++ b/Provenance/Game Library/PVGameLibraryViewController.m
@@ -596,31 +596,22 @@ static NSString *_reuseIdentifier = @"PVGameLibraryCollectionViewCell";
     }
 }
 
-- (NSIndexPath *)indexPathForGameWithMD5Hash:(NSString *)md5Hash
+- (NSArray *)indexPathsForGameWithMD5Hash:(NSString *)md5Hash
 {
-    NSIndexPath *indexPath = nil;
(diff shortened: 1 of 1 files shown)


Provenance/Game Library/PVGameLibraryViewController.m
```diff
@@ -596,31 +596,22 @@ - (void)finishedDownloadingArtworkForURL:(NSString *)url
     }
 }
 
-- (NSIndexPath *)indexPathForGameWithMD5Hash:(NSString *)md5Hash
+- (NSArray *)indexPathsForGameWithMD5Hash:(NSString *)md5Hash
```
leolobato: The same game can show up twice now on the library, once on "recent games" and again on it's own section.

jasarien:
Does this PR depend on #318 ?

leolobato:
Nope, it can be applied by itself. But I just found an issue after renaming -- we need to reload the collection view data in case the game has changed places because of the sorting.

I'll push a fix soon.

jasarien:
alright

leolobato:
OK, ready to merge. It was an issue on both tvOS and iOS. If the game changes position on library after the rename, it would end up duplicated on the list.

PR #360, opened 2016-06-11
Fix app on iOS 8 devices

leolobato:


Provenance/Game Library/PVGameLibraryCollectionViewCell.m
```diff
@@ -104,7 +104,9 @@ - (void)setText:(NSString *)text {
                                                      text:attributedText];
     self.missingArtworkView.image = missingArtworkImage;
     [self setNeedsLayout];
-    [self setNeedsFocusUpdate];
+    if ([self respondsToSelector:@selector(setNeedsFocusUpdate)]) {
```
jasarien: If it compiles with the iOS SDK by itself, then it should be fine. Else, it might need a `#IF TARGET_OS_TV` check.

PR #382, opened 2016-07-27
Game library scrolls very smooth now.

stefz:
Hi,
I worked a little on the Game Library and now the collection view scrolls very smooth. I implemented a NSOperationQueue to download images and I set the max resolution to 200x200 for mobile devices and 400x400 for Apple TV.

I'm also preparing a set of placeholder images for the cartridge without artwork. I already modified the PVGame model to support that. What do you think about that?

Bye,
Stefano

### DIFF
--- a/Provenance/Game Library/PVGame.m
+++ b/Provenance/Game Library/PVGame.m
@@ -7,7 +7,7 @@
 //
 
 #import "PVGame.h"
-
+#import "PVEmulatorConstants.h"
 
 @implementation PVGame
 
@@ -31,6 +31,39 @@
 			 @"originalArtworkURL",
 			 @"md5Hash",
 			 @"systemIdentifier"];
+}
+
+- (NSString *)placeholderImageName {
+    
+    // TODO: to be replaced with the correct system placeholder
+    
+    NSString *systemImageName = @"UNKNONW";
+    
+    if ([self.systemIdentifier isEqualToString:PVGenesisSystemIdentifier]) {
+        return systemImageName = @"GENESIS";
+    } else if ([self.systemIdentifier isEqualToString:PVGameGearSystemIdentifier]) {
+        return systemImageName = @"GAME GEAR";
+    } else if ([self.systemIdentifier isEqualToString:PVMasterSystemSystemIdentifier]) {
+        return systemImageName = @"MASTER SYSTEM";
+    } else if ([self.systemIdentifier isEqualToString:PVSegaCDSystemIdentifier]) {
+        return systemImageName = @"SEGA CD";
+    } else if ([self.systemIdentifier isEqualToString:PVSG1000SystemIdentifier]) {
+        return systemImageName = @"SG 1000";
+    } else if ([self.systemIdentifier isEqualToString:PVSNESSystemIdentifier]) {
+        return systemImageName = @"SNES";
+    } else if ([self.systemIdentifier isEqualToString:PVNESSystemIdentifier]) {
+        return systemImageName = @"NES";
+    } else if ([self.systemIdentifier isEqualToString:PVFDSSystemIdentifier]) {
+        return systemImageName = @"FAMICOM";
+    } else if ([self.systemIdentifier isEqualToString:PVGBASystemIdentifier]) {
+        return systemImageName = @"GAME BOY ADVANCE";
+    } else if ([self.systemIdentifier isEqualToString:PVGBCSystemIdentifier]) {
+        return systemImageName = @"GAME BOY COLOR";;
+    } else if ([self.systemIdentifier isEqualToString:PVGBSystemIdentifier]) {
+        return systemImageName = @"GAME BOY";;
+    }
+
+    return systemImageName;
 }
 
 @end
--- a/Provenance/Game Library/PVGame.m
+++ b/Provenance/Game Library/PVGame.m
@@ -7,7 +7,6 @@
 //
 
 #import "PVGame.h"
-#import "PVEmulatorConstants.h"
 
 @implementation PVGame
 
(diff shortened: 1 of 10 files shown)


appletv4:
Smooth scrolling is a welcome addition , good to see other people with huge libraries 
I have 20gb free space out of 64gb and is mostly all provenance , I'm hoping to use that 20gb for n64 and ps1 in the future 

Stefano any image would be better than a blank white box , in my non coding expertise

Provenance/Game Library/PVGame.m
```diff
@@ -33,4 +33,37 @@ + (NSDictionary *)defaultPropertyValues
...
 }
 
+- (NSString *)placeholderImageName {
+    
+    // TODO: to be replaced with the correct system placeholder
+    
+    NSString *systemImageName = @"UNKNONW";
+    
+    if ([self.systemIdentifier isEqualToString:PVGenesisSystemIdentifier]) {
+        return systemImageName = @"GENESIS";
```
jasarien: Couldn't you use the `shortNameForSystemIdentifier` or `nameForSystemIdentifier` on `PVEmulatorConfiguration` instead of using the literal strings? This would help with localisation in the future.

Provenance/Game Library/PVGame.m
```diff
@@ -33,4 +33,37 @@ + (NSDictionary *)defaultPropertyValues
 			 @"systemIdentifier"];
 }
 
+- (NSString *)placeholderImageName {
+    
+    // TODO: to be replaced with the correct system placeholder
+    
+    NSString *systemImageName = @"UNKNONW";
```
jasarien: Typo here

Provenance/Game Library/PVMediaCache.m
```diff
@@ -85,7 +127,9 @@ + (NSString *)writeImageToDisk:(UIImage *)image withKey:(NSString *)key
         return nil;
     }
     
-	NSData *imageData = UIImagePNGRepresentation(image);
+    UIImage *newImage = [image scaledImageWithMaxResolution:PVThumbnailMaxResolution];
+    
+	NSData *imageData = UIImageJPEGRepresentation(newImage, PVThumbnailQuality);
```
jasarien: Is there a reason for using JPEG here rather than PNG as before? iOS has hardware support for compositing PNGs, but not JPEG.

Provenance/Game Library/PVGameLibraryViewController.m
```diff
@@ -1625,11 +1599,11 @@ - (void)imagePickerController:(UIImagePickerController *)picker didFinishPicking
 	[self dismissViewControllerAnimated:YES completion:NULL];
 	
 	UIImage *image = info[UIImagePickerControllerOriginalImage];
-	image = [image scaledImageWithMaxResolution:200];
+	image = [image scaledImageWithMaxResolution:PVThumbnailMaxResolution];
 	
 	if (image)
 	{
-		NSData *imageData = UIImagePNGRepresentation(image);
+		NSData *imageData = UIImageJPEGRepresentation(image, PVThumbnailQuality);
```
jasarien: Same question here regarding JPEG vs PNG.

Provenance/Game Library/PVGame.m
stefz: Yep.. you're right! I'm sorry but I didn't see these two methods. Obviously, this is the best way. And so the "UNKNOWN" type is useless.

Provenance/Game Library/PVMediaCache.m
stefz: Just because downloaded uncompressed images are about 1,5MB and jpeg/resized versions are around 100kb. This is why the "collectionView" it wasn't smooth.

jasarien: If you resize them, but still use the PNG representation, is the collection view still slow to scroll? What is the size of a PNG vs a JPEG at the scaled  down size?

Provenance/Game Library/PVGame.m
Armored-Dragon: I would still include an "UNKNOWN" type, just in case things somehow go wrong.

Provenance/Game Library/PVMediaCache.m
Armored-Dragon: Usually, JPEG is advantageous, since boxart is usually like a photo. For some boxarts, PNG will be better as it is better for cartoony scenes. For gradients, JPEG is better for compression ratio.

Offtopic: Could you invite me to Slack?

jasarien: DM me your email on twitter or something.

Provenance/Game Library/PVGame.m
stefz: I think the "UNKNOWN" type is already checked in ```shortNameForSystemIdentifier``` and ```nameForSystemIdentifier```. If not, it's a good idea to include it.

Provenance/Game Library/PVMediaCache.m
stefz: I'll check the different weights of JPG/PNG resized images asap

Provenance/Game Library/PVGame.m
jasarien: `shortNameForSystemIdentifier` and `nameForSystemIdentifier` will just return the values for those keys in the Systems Plist. They won't do anything special for an unknown case. They'll just return `nil` if you pass an unknown system ID.

Provenance/Game Library/PVMediaCache.m
stefz: JPG scaled around 10-20KB, PNG scaled around 80-100KB. In both cases collection view is very smooth. We can use the PNG scaled.

stefz: Done

Provenance/Game Library/PVGame.m
stefz: Done

PR #396, opened 2016-08-27
Snes9x update

willco007:
Includes previous pull request plus Snes9x core update.


jasarien:
Doesn't this pull request contain all the same content as your previous PR, with the addition of the SNES9x update?

If that is the case, you could close your previous PR and retain this one.

willco007:
Yup, sure thing.

willco007:
I've applied the changes from your comments on the FCEU PR as well as removed the background image border on the tvOS artwork as the OS applies it automatically.

Provenance.xcodeproj/project.pbxproj
```diff
@@ -1247,7 +1247,7 @@
 					};
 					1AD481B31BA350A400FDA50A = {
 						CreatedOnToolsVersion = 7.1;
-						DevelopmentTeam = MF85NRPL4W;
+						DevelopmentTeam = 63497P68S6;
```
leolobato: It would be best to not change the development team.

willco007: Yes, this commit is to change it back to what it should be.

jasarien:
It seems theres probably some issues with this SNES update: #417 Could you check them out @willco007 ?

PR #405, opened 2016-09-17
Add basic gameplay recording

marcelvoss:
Added a simple option to the emulation menu that lets users record their gameplay using ReplayKit. Here's an [example recording](https://www.dropbox.com/s/zf7985auypov8w1/IMG_0052.MP4?dl=0).


marcelvoss:
Will submit a new PR at a later point.

PR #413, opened 2016-09-22
Added support for iCade 8Bitdo SFC39 Gamepad

rafcabezas:
This probably also works for the 8Bitdo SNES30 Gamepad


rafcabezas:
Found an issue -- correcting.

PR #414, opened 2016-09-22
Added support for iCade 8Bitdo SFC30 Controller

rafcabezas:
This probably also works for the 8Bitdo SNES30 Gamepad


marcelvoss:
If so we should probably merge both 8Bitdo controllers into a single class.

rafcabezas:
Good point, I'll work on that tomorrow. Feel free to merge this pull request in the mean time. I'll push a new one tomorrow, or as soon as I have free time.

appletv4:
Will this work on tvOS or iOS only ?

rafcabezas:
It should work on tvOS as well. But not tested.

rafcabezas:
opening new one with the changes.

leolobato:
tvOS doesn't support iCade controllers (including 8bitdo), unfortunately.

rafcabezas:
Gotcha. Makes sense, actually.

Issue #144, opened 2015-10-13
Toggle for Hiding Title Labels in Game Library


ghost:
For a cleaner look, a toggle to hide title labels in the Game Library screen would be nice. If there is no artwork, it can be displayed where the artwork would have been.

marcelvoss:
Something like this?
![simulator screen shot 20 09 2016 22 55 28](https://cloud.githubusercontent.com/assets/5435925/18688597/5db6676a-7f85-11e6-835f-e491cd7a65fb.png)

Armored-Dragon:
Could you make a pull request or something for that?

marcelvoss:
I'm improving it at the moment and fixing small bugs that I caused due to this. Will submit a PR as soon as it's done.

marcelvoss:
See pull request #415 for progress regarding this issue.

marcelvoss:
Merged with #415. Can be closed.

PR #415, opened 2016-09-22
Implement #144: hide game title labels

marcelvoss:
- add new appearance view controller and a corresponding menu in settings
- add switch for hiding game titles in game library

Notice: Due to new placeholder system, we may need to find a way to display the title of the game on the image otherwise it will only show the placeholder but without any information regarding the game itself.


marcelvoss:
Here's a preview of the slightly changed UI:

rafcabezas:
Love it!

PR #416, opened 2016-09-22
Added support for iCade 8Bitdo SFC30 Controller

rafcabezas:
This probably also works for the 8Bitdo SNES30 Gamepad ~~and tvOS~~.
Also, now integrated into a single class as requested. (see #414)


rafcabezas:
Feedback would be appreciated by users / devs / project owner. Thanks!

tiagoln:
I bought an 8Bitdo SNES30, I'll give some feedback when it arrives.

rafcabezas:
Weird, you're right. When I first tried the SFC30 the mappings were all weird. This pull request is not needed.

PR #419, opened 2016-09-25
Add screenshots option

marcelvoss:
You can now take screenshots of the emulation without seeing the controls etc.


marcelvoss:
Two quick examples of the screenshot quality:
photo 25-09-2016 13 05 49
photo 25-09-2016 17 55 48

jasarien:
Works fine, but it also capture the FPS counter if it's enabled, because the FPS label is added as a subview to the GL view.

marcelvoss:
Fixed with latest commit. The FPS label will now fade out before taking a screenshot and will fade in again after the screenshot has been taken. Since showing the FPS is mostly a debug option this should be fine.

PR #420, opened 2016-09-26
Fix tvOS build

drakkhen:
Needed to add/tweak some preprocessor conditionals.


james-rantmedia:
Thanks for this 👍

jasarien:
Thanks for this 👍

PR #426, opened 2016-10-01
Show current and master SHA hashes in Settings

drakkhen:
Adds new Build Phases to create local.plist (which is ignored by git)
containing the SHA of the current branch's most recent commit and the SHA
of the commit in master from which the local branch was created. The latter
updates after rebasing against an updated master branch.

Both of these SHAs can then be seen on the Settings menu, which may be
helpful for both developers and users (who may be having issues building and
installing Provenance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@ build
 .svn
 xcuserdata
 *.mode1v3
-
+Provenance/Resources/local.plist
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@ build
 .svn
 xcuserdata
 *.mode1v3
-Provenance/Resources/local.plist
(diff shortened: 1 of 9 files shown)


drakkhen:
Two things I'd like to bring to attention with this MR:

- because the new local.plist file is transient to the build process, sometimes it is not present on disk before building and then it shows up red in the Project Navigator at Provenance/Resources/local.plist.  Not sure if that's a concern (it looks like something's wrong, even though it's fine) or if there is any way to workaround that.
- I tried making the smallest changes necessary to Storyboards -- just needed to add two new Table Cell Views -- but the diff is showing a bunch of seemingly unrelated small changes (replacing height="49" with height="49.5" for example).  This is the first time I've worked on an iOS/tvOS project so I don't know if that's expected or if I should go back and fix those unexpected changes by hand.

Thanks!

JoeMatt:
Instead of making a new plist at build, you can add a build script phase to modify the Info.plist after the copy phase. This leaves the original Info.plist untouched so you don't have to commit it after each build (Which is how we should be doing build version). You can either 'sed' it like you did, or use plistbuddy command line tool to add/change plist keys. You would use a path like $BUILT_PRODUCTS_DIR/Info.plist i believe. As long as it's after the copy phase, and before the build phase where code-sign runs, it's ok to modify anything in the build directory.

Here's a gist that I wrote for one of my other projects,
https://gist.github.com/JoeMatt/2ae4e57f09bd19e15314f32f1c02e880

drakkhen:
OK cool.  Yeah I was on the fence about maintaining the new values alongside the existing build information or putting it in a separate file that could be ignored by git.

drakkhen:
Rebased against master

Issue #435, opened 2016-10-15
Output is not precisely pixel aligned on iPhone 7


msikma:
On iPhone 7 (not sure if it's the same on iPhone 6) the output is not exactly scaled to an integer ratio, and there's no smoothing, so there's a bit of distortion in the sense that some pixels are displayed larger than others.

![slack for ios upload](https://cloud.githubusercontent.com/assets/1507745/19413663/bd11e97a-9333-11e6-8a2f-3be7c07523ae.jpg)

In this picture you can see that the ground tile has some of the pixels from the game's tileset displayed as 3x3, 3x2, 2x3 and 2x2.

The iPhone 7's display (in portrait) is 750x1334, the NES is 256x224, and 750/256 = ~2.93. It's almost 3, but the difference is large enough to be noticeable in both still pictures and in motion.

I think the solution is to allow an option to scale the pixel art to `floor(screen width / console width)` and then optionally also scale up to the actual screen width with smoothing. This way people could choose between nearest neighbor scale to full width, nearest integer scale (x2 on portrait), or nearest neighbor to x2 and then bilinear scale to full width.

PR #437, opened 2016-10-16
Image smoothing option

msikma:
This is my first ever Obj-C PR, so some things might be wrong. Please let me know if that's the case!

This adds an extra switch to the settings menu (default state is off) which, when activated, enables bilinear filtering for the output instead of nearest neighbor. I think this is a nice solution to #435 which I raised earlier.


msikma:
[Screenshot of the new setting](https://cloud.githubusercontent.com/assets/1507745/19417517/5add34d6-93af-11e6-9841-8f878220d33d.PNG), and [screenshot of the result](https://cloud.githubusercontent.com/assets/1507745/19417518/5addd2f6-93af-11e6-96be-609756eae15a.PNG).

PR #445, opened 2016-10-21
Fixes iPhone 7 / 7 Plus random crashes

rafcabezas:
…the past

### DIFF
--- a/PVSupport/PVSupport/OETimingUtils.m
+++ b/PVSupport/PVSupport/OETimingUtils.m
@@ -53,7 +53,16 @@ void OEWaitUntil(NSTimeInterval time)
 {
     init_mach_time();
 
-    mach_wait_until(time / mach_to_sec);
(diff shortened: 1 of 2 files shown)


lopeza511:
Seems to be working fine. I do get some lag every now and then. I am not sure if this was an issue before.

rafcabezas:
@lopeza511 If you have a chance, please re-test (I moved the logic / refactoring).
Yes, I do see the lag; I believe this lag was causing the crashes before, now the game continues playing. I'm not sure if the lag exists on devices != iPhone7. But it might be worth some profiling. Definitely a different (but related) issue, though. (That is, this fix is not introducing this lag, rather, preventing it from crashing the phone).

jasarien:
Seems to work ok as a temporary workaround for the issue at hand. I'll merge it while we continue looking for a more permanent solution.

rafcabezas:
Sounds like a good plan!

PR #453, opened 2016-10-27
New, simpler time-accurate emulation loop thread.

rafcabezas:
Removed unused OETimingUtils.


jasarien:
Does this supersede your other PR regarding the crashes on iPhone 7?

rafcabezas:
In my opinion, yes. But it's up to you if you want to merge the other one, since it's way less invasive. 
I have been using this branch for over a week without issue, though. Your thoughts welcome.

rafcabezas:
BTW, I did some quick profiling to see why the iphone7 was lagging (the root cause of the crash); and I can see in Instruments, where sometimes CPU usage goes to 100%. This is very strange, and only seem to affect the iPhone 7 (models). I don't know what's causing this yet.

jasarien:
Tested this today and I've seen a few issues.

It seemed OK in SNES emulation with Chrono Trigger. However Genesis emulation with Sonic 1 showed up some flaws: the overall frame execution was slower than it should be, and when the game time was out of sync it 'fast-forwards' to catch up. Overall, I think this change has impacted Genesis emulation in a pretty bad way and probably affects other cores too.

This was my gut feeling  about leaving the mach_timing APIs behind -- I'm not sure what other options there are for timing as accurate as we need it to be for emulation...

rafcabezas:
Interesting. I'll take a look at the Genesis core see if I can recreate. But I do think that both this and PR 445 work-around the real issue with iPhone 7: random slowdowns. We need to figure out what's creating those, because that's the real root-cause.

jasarien:
@rafcabezas some more conflicts here too

rafcabezas:
Yeah, needs to be rebased after merging Joe's cores. 
I'll work on it asap.

rafcabezas:
If you agree, @jasarien, I'm gonna close this PR, and instead create individual PRs for some good improvements that I used here.

jasarien:
Gave this another test today, was working really well. No stuttering issues on my iPhone 7. Games ran smoothly and framerate held up. I'm going to merge this in. If people have issues we can consider reverting it, but from what I've seen it should be OK.

rafcabezas:
Nice!

PR #456, opened 2016-10-28
TV cell sizing

davidmuzi:
### What this change is

This change generally makes the cell sizing larger on tvOS.  This uses the recommended sizing for a 5-column collection view as recommended in the [HIG](https://developer.apple.com/tvos/human-interface-guidelines/visual-design/)
### How it does it

The HIG recommends a cell width of 308 pix with 50 pix horizontal spacing.  This is up from 250 width.  This change also introduces an aspect ration per console type, grouped into 3 groups: wide, tall and square.  This allows a cell height determined by the aspect ration of the box art.  This greatly decreased the white vertical space between cells on landscape boxart, like SNES.
### What it looks like

Before
simulator screen shot oct 28 2016 5 56 11 pm

After
simulator screen shot oct 27 2016 9 53 55 pm


jasarien:
I noticed you added some commits since Leo last commented. Is this complete now?

davidmuzi:
Yes.  Id like to have some more testers give it a try with varying consoles and library sizes.  Ive tested with NES, Genesis, and SNES.

iOS should be unaffected.

jasarien:
I'll merge it in -- if there's any issues, I'm sure people will let us know. They seem to like doing then :P

RedLucas:
For what it's worth, this small design change looks really great.

appletv4:
i reinstall provenance a few hours ago, version 1.27(508), is this the new version after jasarian added this new pull ?

PR #485, opened 2016-12-20
Favorite Games

mattsson:
Hi.

I'm not sure you even want this feature, but I thought I'd make a pull request just in case. I added a "Toggle Favorite" option to the action sheet, making it possible to set a list of favorite games, which will be shown in their own section, on top of Recent Games. I have thousands of games in the app, and this feature helps me out a lot.


appletv4:
This is useful to me thanks , I have like 10gbs of roms in provenance 
This is for both iOS and tvOS ?

mattsson:
I have not tested this on tvOS, but it should work there as well.

Issue #286, opened 2016-02-15
Add ability to change cover art via web server's UI


drakkhen:
It would be very convenient if the web server's UI showed current cover art which, when clicked, allowed users to upload alternative art.

This would also solve the repeated requests to change artwork on tvOS (#173, #188, #220, #225, #237, #284).  I realize that the tvOS interface is not an ideal (or even possible?) medium for feature.  Adding it to the web server could provide the missing tvOS functionality as well as an easier UI to work with on other devices.

Personally, I'd be okay with dropping this functionality from the iOS UI altogether and just having the feature on the web UI so there is only one way to do it, and one feature to maintain.

Hanterdro:
+1

stuartjmoore:
If anyone has any suggestions of where to start, this is something I’m going to try and implement.

jasarien:
Not sure where you'd start - the main thing that needs to happen, though, is setting the property on the Realm-managed Game object for the artwork file url, which I'm not sure how you'd do via the web UI... 

It's not as simple as dumping the image into a folder somewhere and having the game find it automatically.

lolkthxbai:
What about having a folder where you could upload custom artwork and then in the settings, assign the artwork to the correct game name?

jasarien:
@davidmuzi was working on a way of adding artwork on tvOS using an image search UI.

lolkthxbai:
Do you know if it searched for images online, a database of game covers or a local folder?

Also, how could I join the slack group?

jasarien:
DM me your email on twitter (@jasarien)

jasarien:
I don't know the specifics on the search. I think it might have just been something like a Google image search

stuartjmoore:
@jasarien My going theory (still new to this codebase) was to mimic `moveCDROMToAppropriateSubfolder:` and import the images into the cache based on their filename.

stuartjmoore:
@jasarien Finished (cleaning up for PR): https://github.com/jasarien/Provenance/compare/master...stuartjmoore:issue/286-web-ui-cover-art

It’s pretty basic. Name the image file the same as the rom file (`Super Mario Brothers 1.nes` → `Super Mario Brothers 1.nes.png`) and drag-and-drop into `/Cover Art`.

lolkthxbai:
[...]

PR #490, opened 2016-12-24
Ability to Change Cover Art via Web Server’s UI

stuartjmoore:
Adds a directory watcher, similar to the ROMs directory watcher, that will import image file to matching game cover art.

> You can upload new cover art images by uploading a *.zip file into the Cover Art/ directory. The images need to be named the same as their ROM (IE, “kart fighter.nes.png” will update the artwork on “kart fighter.nes”. ROM Filenames can be found in the “com.provenance.*” directories). Then close the web server and Provenance will attempt to import them.

Issue #286

### DIFF
--- a/Provenance/Game Library/PVGameImporter.h
+++ b/Provenance/Game Library/PVGameImporter.h
@@ -13,6 +13,8 @@ typedef void (^PVGameImporterCompletionHandler)(BOOL encounteredConflicts);
 typedef void (^PVGameImporterFinishedImportingGameHandler)(NSString *md5Hash);
 typedef void (^PVGameImporterFinishedGettingArtworkHandler)(NSString *artworkURL);
 
+@class PVGame;
+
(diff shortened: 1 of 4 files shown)


Hanterdro:
Ehen this will be merged?

jasarien:
When I get a chance

lolkthxbai:
I've been testing this out for a while now and it works great. I was able to upload the photo to the directory and have provenance immediately detect and assign the image to the correct rom. I've only tried this with SNES roms but so far it's work great for all my roms missing their cover art.

gareauk:
I've tried to do this and can't seem to import any covert art.  I created a folder called "Cover Art" and I renamed the images as described above and still a no go.

I'm using iOS 10.2.1 on a iPad Mini 4.

jasarien:
It's not in the build you're using, @gareauk

PR #492, opened 2016-12-27
Add new properties to PVGame

marcelvoss:
We now parse the game description, its release date, its developer and publisher as well as its genre.

This is useful for Spotlight indexing or new 3D Touch functionalities.


marcelvoss:
See #494 which includes this PR as well.

PR #494, opened 2016-12-28
Add URL scheme

marcelvoss:
Adds ability to open any game from anywhere on the platform using the following scheme: provenance://play?md5=xxxxxxxx

It requires the MD5 hash for every game.


Provenance/Provenance-Info.plist
```diff
@@ -37,8 +37,19 @@
...
+			<string>Editor</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>provenance</string>
+			</array>
+		</dict>
+	</array>
 	<key>CFBundleVersion</key>
-	<string>632</string>
+	<string>639</string>
```
jasarien: Try not to commit changes to the build number and Revision number (below) since these change on build-by-build basis, they're likely to cause conflicts with other PRs.

marcelvoss: Oops, sorry. 🙃

Issue #311, opened 2016-03-22
Can't access Saved States using Menu button in tvOS 9.2


KrisasMan:
Since I upgraded to tvOS 9.2 I can no longer access saved states by pressing the Menu button on my controller or Siri remote, it just goes back to the games library instead of opening the saved state menu.

I have re-synced my code using terminal 
`git pull origin master`
and upgraded my Xcode to 7.3 and re-built to tvOS 9.2 but still no change in behavior.

jeaz:
Have the same issue.

CodelabOne:
Same on my side. Menu button works like the back button now.

jasarien:
I had a look at this bug this evening and I can't figure it out...

I can see that the `pauseButtonHandler` is called, but before any of its implementation is executed, the `emulatorViewController` is dismissed...

I can't figure out why it dismisses the view controller when `controllerUserInteraction` is set to false. Anyone have any ideas?

arbitrarytech:
Experiencing the same issue using my Nimbus Controller (running latest firmware v1.6.4.1) since the 9.2 tvOS update.

madangus:
Same with the hori controller.

Great project team.

Hopefully it ends up being easy. Trying to see if there's a way to force downgrade of tvos in meantime.

jonwills:
Are battery saves still working for you?  I am experiencing the above issue and battery saves don't seem to be working.

binarydev:
Having the same issue (Nimbus), although it seems like a general issue that will happen with any controller. Tried tracking logical flow on button press, but in the emulator it works perfectly every time. On device, I'm not hitting any of my breakpoints. So much for family retro game night this weekend :disappointed:

djsyche:
Id like to report the same thing. I actually do not know if it would have worked on my device prior to the upgrade as I was not using Prov previous to the 9.2 upgrade. I finished installing it today and got about an hour of zelda lttp in and saved after returning to my game play I noticed that I actually had no way of using saved states or bat saves. Its sort of weird. But essentially the saves are there in their appropriate folders and I am just not able to access any of them. In regards to the Menu button I am being booted directly to the app landing screen with the rom list laid out and no options present.  Some times upon exiting the back ground music will continue. This has happened with all 30 or so different roms I have tried to use.

leolobato:
[...]

PR #319, opened 2016-03-28
Fix menu button during gameplay on tvOS 9.2

leolobato:
Use a tap gesture recognizer since it will override the default tvOS
'back' functionality.
Resolves #311

PR #503, opened 2017-01-02
Fix Issue #311 by ensuring that gestures for Menu presses on tvOS are…

carloperez:
… caught and handled appropriately.

This should allow you to press the 'Menu' button on both the controller and the remote from the game menu without being kicked back out to the TopShelf view.  Smoke testing of various areas of app using menu button shows this works well and kicks you back to navigator then springboard from TopShelf view as expected.

Warning: Tested only on 10.1 with a SteelSeries Nimbus controller and the stock AppleTV 4 remote.

### DIFF
--- a/Provenance/Emulator/PVEmulatorViewController.m
+++ b/Provenance/Emulator/PVEmulatorViewController.m
@@ -805,6 +805,20 @@ void uncaughtExceptionHandler(NSException *exception)
 
 #pragma mark - Controllers
 
+#if TARGET_OS_TV
+// Ensure that override of menu gesture is caught and handled properly for tvOS
+-(void)pressesBegan:(NSSet<UIPress *> *)presses withEvent:(nullable UIPressesEvent *)event {
+    
+    UIPress *press = (UIPress *)presses.anyObject;
+    if ( press && press.type == UIPressTypeMenu && !self.isShowingMenu )
+    {
+        [self controllerPauseButtonPressed];
+    }
+    else
+        [super pressesBegan:presses withEvent:event];
+}
+#endif 
+
 - (void)controllerPauseButtonPressed
 {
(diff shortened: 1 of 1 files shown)


carloperez:
Related to #319 and #311.

@leolobato 
@jasarien

jasarien:
Does this solve #506 as well? (Seems that perhaps this bug was reintroduced when the latest icade+tvOS code was merged recently).

maddox:
I'm taken directly back to the home screen since this change. The only way to avoid is commenting out:

```objc
#if TARGET_OS_TV
// Ensure that override of menu gesture is caught and handled properly for tvOS
-(void)pressesBegan:(NSSet<UIPress *> *)presses withEvent:(nullable UIPressesEvent *)event {
    
    UIPress *press = (UIPress *)presses.anyObject;
    if ( press && press.type == UIPressTypeMenu && !self.isShowingMenu )
    {
//         [self controllerPauseButtonPressed];
    }
    else
    {
        [super pressesBegan:presses withEvent:event];
    }
}
#endif
```

Which is weird, because I can't even get a break point to work inside it :/

jasarien:
I did exactly this yesterday. https://github.com/jasarien/Provenance/commit/b778e1a0d5b0bd108461414184f8852047a830cb

The latest source should work as it basically does the same as what you mentioned: comments out the line. Lol

maddox:
Hmm. I was on `master`, but I'll try pulling again. Thanks!

maddox:
Weirddddd. that commit IS on my branch and it's still doing it :/

jasarien:
Tried a clean build?

maddox:
yeah, in fact, i was incorrect in saying commenting out that entire portion prevented it, it still happened. I'm not sure what I was futzing with that got it to work as normal :/

I'll look at it some more and report back.

PR #508, opened 2017-01-06
Fix TopShelf

davidmuzi:
### Problem
The realm version  was bumped in the app target, and not in the Top Shelf extension.  

### Fix
Extract Realm config to single place to prevent multiple targets from using different configs


carloperez:
Cherry picked this and seems to be working.  Saw some odd behaviour when I yanked the USB-C cable out (kicked to springboard and wouldn't immediately load from top shelf) but started working a few tries after.  Seems to be working well aside from that anomaly!

appletv4:
Paid developer account still needed for this @davidmuzi ? No way to make it work for free 7 day accounts ?

jasarien:
The limitation is on AppGroups. They're required for TopShelf functionality. I'm not 100% clear on the restrictions of free accounts, but if AppGroups aren't available then, yes, you'll need a paid account.

appletv4:
I'm checking now and is letting me  checkmark an app group without any errors , i am using a new bundle Id name that expires on the 12th
i think it will work this time with a free account , i will test when you merge this to the official provenance  and report back 

in the past i was using an expired bundle id maybe thats why it didn't work
http://mybyways.com/blog/new-limitations-imposed-on-free-apple-developer-account

appletv4:
update i couldn't get this working even without getting any errors with app groups on and my own group id 
only an image shows up in top shelf and i don't see a setting in the app ( don't remember if there was one? ) to turn it on

"group.provenance-emu.provenance"


do i only remove provenance-emu or provenance-emu.provenance  ? when adding my own group id
i was doing it the second way

jasarien:
You can use anything as along as it's unique, so just changing the `provenance-emu` part is fine. There never was a setting - it's supposed to Just Work™.

jasarien:
@davidmuzi just tested this and it works great, nice work. One question: when I merge this, should I commit it with App Groups initially turned off, to prevent people with free accounts getting build errors?

davidmuzi:
Good point, yea default it to off

David Muzi

On Fri, Jan 06, 2017 at 4:50 PM James Addyman

<
mailto:James Addyman <notifications@github.com>
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

https://github.com/davidmuzi
just tested this and it works great, nice work. One question: when I merge this, should I commit it with App Groups initially turned off, to prevent people with free accounts getting build errors?

—

You are receiving this because you were mentioned.

Reply to this email directly,
https://github.com/jasarien/Provenance/pull/508#issuecomment-271017618
, or

PR #543, opened 2017-03-02
D-Pad stays pressed with touchDown outside until touchup + Disable D-pad Diagonals Option

nilsleiffischer:
When pressing the D-Pad and then dragging outside its area, the direction now remains pressed. I found this more convenient to play, since without physical buttons the control area is easy to leave with the thumb.

I also added the option to disable diagonal directions of the D-Pad on systems that do not support it even on the physical device, such as the Gameboy Advance. For this I added the `PVControlDiagonalDirectionsEnabled` key in `PVEmulatorConstants.h`. The key can be added with a boolean value to any dictionary of `PVDPad` control type in the `systems.plist` that describes the emulated devices. So far I only added it to the Gameboy Advance.

Both changes make it more natural to play on Gameboy Advance, I found. Any comments on this?

### DIFF
--- a/Provenance/Resources/systems.plist
+++ b/Provenance/Resources/systems.plist
@@ -447,6 +447,8 @@
 				<string>PVDPad</string>
 				<key>PVControlSize</key>
 				<string>{180,180}</string>
+				<key>PVControlDiagonalDirectionsEnabled</key>
+				<false/>
 			</dict>
 			<dict>
 				<key>PVControlType</key>
--- a/Provenance/Controller/JSDPad.h
+++ b/Provenance/Controller/JSDPad.h
@@ -33,6 +33,7 @@ typedef NS_ENUM(NSUInteger, JSDPadDirection)
 @interface JSDPad : UIView
 
 @property (nonatomic, weak) IBOutlet id <JSDPadDelegate> delegate;
+@property (nonatomic) BOOL diagonalDirectionsEnabled;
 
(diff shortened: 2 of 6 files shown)


Provenance/Resources/systems.plist
```diff
@@ -447,6 +447,8 @@
 				<string>PVDPad</string>
 				<key>PVControlSize</key>
 				<string>{180,180}</string>
+				<key>PVControlDiagonalDirectionsEnabled</key>
```
jasarien: It looks like you’ve only added this to the GBA system and none of the others.

nilsleiffischer: Yes, I wrote so above. We can also add the key to other systems that have no physical D-Pad buttons in diagonal directions.

sevdestruct:
@nilsleiffischer are you going to complete this feature? It is now quite a ways behind the current dev branches, you might want to get this in before the dev branches get merged into master for 1.5, as the code base is switching to swift at that point…

Provenance/Resources/systems.plist
sevdestruct: A lot of games allowed pressing both axis simultaneously to get a diagonal axis regardless of the lack of a physical diagonal, just sayin'.

JoeMatt:
In the most recent dev branches controller definitions and PVControllerViewController have been completely rewritten so this PR won't merge at all. If this feature is requested was can implemented in the new code base pretty easily.

PR #561, opened 2017-04-07
Add Mednafen port

JoeMatt:
Support for,

- Atari Lynx
- Neo Geo Pocket (Color)
- WonderSwan
- Nintendo Virtual Boy
- PC Engine/TurboGrafx 16 (CD)
- SuperGrafx
- PC-FX
- Sony PlayStation


jasarien:
👍 Awesome work @JoeMatt.

The more people we can get to test this PR out the better.

JoeMatt:
In addition to Mednafen, I added 7z support since most PSX iso's I found were in that format.
Also the UI for conflicts on ATV wasn't hooked up. Fixed that, though it's somewhat unstable when multiple .bin's are in a package.

XeresRazor:
There's a few places where I find it hangs. One is after importing PSX games via the web interface. The "Extracting roms" dialogue never dismisses, and I have to kill and restart the app to get the conflict resolver.
Also the conflict resolver itself cannot be dismissed and requires a restart of the app. I've worked around it in my local copy by dismissing when there are no more titles to resolve by adding the following code to the end of -updateConflictedFiles in PVConflictViewController.
```if (self.conflictedFiles.count == 0) {
        [self.navigationController dismissViewControllerAnimated:YES completion:^{
            
        }];
    }```
We should implement an actual back button to dismiss however.

XeresRazor:
I think I've found the source of the extraction hang, in `- (void)extractArchiveAtPath:(NSString *)filePath` `-stopMonitoring` is called, which sets the `extractionCompleteHandler` to NULL, thus preventing it from running and dismissing the hud when extraction is completed.

XeresRazor:
It looks like Playstation import doesn't work correctly, only .cue files are moved after extraction, so a bunch of .bin files get left in Documents/conflict and throw an assertion when they can't be found when loading a game. Current workaround is to move the files via the web interface.

JoeMatt:
@XeresRazor The import issues should be resolved in #566 I beleive

PR #563, opened 2017-04-10
Pokemon Mini emulator

JoeMatt:
A rare [portable device](https://en.wikipedia.org/wiki/Pokémon_Mini). ROMs end with .min

Tested with onscreen controls.
MFi might work. Untested.


JoeMatt:
Merged into #566

PR #564, opened 2017-04-12
Add Atari 800 core

JoeMatt:
Atari 800 core adds support for Atari 5200.

Does not currently support other Atari 8bit systems since they are 'PC' type systems that use a keyboard and mouse.

Does not support number pad controller input for Atari 5200 but this isn't a requirement for all (most?) games.

Requires BIOS 5200.rom. Extension is .a52. Most roms I ran across were named .bin so I simply renamed them. I'd rather not muck up the already overused .bin extension until we have md5 rom matching for conflicts.


JoeMatt:
Merged into #566

hstriepe:
I just added an a52 ROM, but r683 does not recognize it. What do I do to get it into action.
Is there a way to make the 800 stuff work? It was my first comp.

Issue #506, opened 2017-01-04
Menu button on Steelseries Nimbus controller bypasses in-game menu, goes straight to games library view


starlessblack:
So I just downloaded and built ver 1.3.1 a few days ago, and I'm having this problem: SteelSeries Nimbus controller, ATV4 running latest iOS ver.10.1--Menu button on Nimbus controller goes straight back to Provenance games library, not to in-game menu. My workaround has been to quickly double-tap the Siri button on the ATV remote which briefly brings up the Siri listening dialog but also has the side effect of pausing the game and bringing up the in-game menu.

I also noticed that if I turned on the iCade controller support in Provenance settings, it appears to fix the menu button issue, however, it also appears to map the Nimbus' "B" button to Menu. Or, rather, it's double-mapped--if you hit the B button, you can briefly see your in-game sprite or whatever start doing what it's supposed to do (jump, shoot, stab, run, flee, fly, kick, dance, etc.), but only for a brief millisecond before the in-game menu pops up.

rogersbowne:
I am having the same exact issues in the same situations.

OKNoah:
Affects Mad Catz CTRLi too.

jasarien:
I just merged #503 - you can try the latest code in master and see if it fixes the issue?

starlessblack:
Yep! That #503 fix did the trick! Thank you, and keep up the great work :-)

zfisher323:
First. Great app, thanks! Just recently updated from 1.2.7 to 1.3.2. I'm having the B button issue. Using nimbus steel series controller. When I press the B button in-game, it does the B butting Action but also immediately pulls up the provenance in-game menu. Am I doing something wrong? Thanks. Zack

zelik:
Same issue as zfisher,  don't know if this is something wrongly configured on my side?  Love this app. Helps with being in hospital!

lolkthxbai:
@zelik the issue was resolved but resurfaced again PR #561 resolves the issue but was merged into PR #566 As soon as the pull request is merged into master you can update the app to fix it. In the meantime, using the Siri button tvOS can be used as a workaround

Issue #251, opened 2015-12-21
Scanlines & CRT filters


sakuranbokun:
I've been playing a good amount of OpenEmu lately, and it has a nice selection of filter effects. Some of which add scanlines and some very convincing CRT style blur, colour bleed, and phosphor glow effects. It looks remarkably like a real CRT.

I realise OpenEmu is running on a Mac, and so has power to burn, but would it be possible to replicate any CRT style effects in Provenance? Crisp and blocky is nice, but fuzzy, glowy, scan line covered CRT is proper old school.

pmp174:
#54  is already opened for this

jasarien:
Correct, dupe of #54

Issue #56, opened 2015-03-13
Button mapping customisation for iOS 7 game controllers.


jasarien:
Allow custom button mappings for physical controllers.

appletv4:
on nes and the nimbus controller is hard playing super mario when A is run and B is jump
[img]http://i.imgur.com/HKfQncl.png[/img]

it should be X to run and A to jump

jonathantneal:
You found your way here because you bought a Steelseries Nimbus and were surprised by the NES button mappings in Provenence.

![Steelseries Nimbus](https://cloud.githubusercontent.com/assets/188426/11453943/919ee400-95eb-11e5-85a1-c16284ad970b.png)

You expected the Nimbus X button to map to the NES B button, and the Nimbus Y button to map to the NES A button.

![image](https://cloud.githubusercontent.com/assets/188426/11453948/bd95c16e-95eb-11e5-89e1-68ff882f2b11.png)

Instead, the Nimbus A button maps to the NES B button. and the Nimbus B button maps to the NES A button.

> That's by design. Start and select had to go some where. On a standard controller without triggers, the only buttons not in use on NES were the X and Y buttons. So I assigned them to start and select... you can wait until the customizable controller mapping is implemented

Here are two solutions:

1. Wait until a button mapping feature is made available. Be really thankful for this software in its current form, and applaud @jasarien for his efforts.

2. Figure out how to change the mappings yourself and share the solution here. I’ve tried playing with `PVNESControllerViewController.m`, myself, which has not worked. Good luck, fellow novices. :)

appletv4:
Or 3. download this which I have been using 
https://github.com/nlflint/Provenance.git

OutRunner:
I don't know if should submit this as a separate enhancement request, but would it be possible to add turbo-fire to the possible button mappings?

So for example in Contra, instead of mapping NES B to Gamepad A, I could map NES Turbo B to Gamepad A. 

This does possibly open up the need for per-game mappings though as you probably wouldn't want to keep remapping the controls every time you switch between say Mario and Contra.

MysticalOS:
[...]

PR #566, opened 2017-04-14
Mednafen+atari800+poke

JoeMatt:
This PR combines #561 #563 and #564 in addition to some bug fixes, and warning fixes.


XeresRazor:
It doesn't look like any of the new cores were included in the .xcworkspace. It fails to build after a checkout and configure, and needed those projects added first so that Xcode knew what to build.

XeresRazor:
After importing a Playstation CD, and resolving the conflict, the .cue file is moved, but none of the associated .bin files are relocated.

JoeMatt:
@XeresRazor You have to close the workspace and re-open when switching branches. If though Xcode asks to reload the workspace from disk it doesn't load it correctly when subprojects change. It's a bug in Xcode, happens on my other projects.

What are the file names of the bin/cue pair?

XeresRazor:
Oh, duh, I should have remembered that, been stuck in non-Xcodeland too long.
The bin/cue set are Ridge Racer (USA).cue and Ridge Racer (USA) (Track 01).bin - Ridge Racer (USA) (Track 14).bin

XeresRazor:
Playing a PSX game, exiting to the Library, and then playing another game, results in the second game running at half speed or less. Terminating and relaunching Provenance resolves.

JoeMatt:
@XeresRazor An easier way to resolve that is hit the menu butt, choose speed settings and choose normal.

It's technically already in normal mode, but re-selecting normal causes from frame rate calculations to reset. Sometimes they initialize incorrectly. It's a known issue we're working on.

Cepheidsong:
The BIOS are off in this version for PSX and Sega CD both.  It wants the SCPH for the PSX in lower case than they usually come, and wants the Sega CD "bios" part of the bin files in capitals.  This causes the Sega CD roms not to work, and I don't know if it's making the psx choppy, but it's really choppy even after adjusting speed.

This problem is not present on the mednafen-only branch, or the current master/normal branch.

carloperez:
I haven't dug so there might be a really good reason, but could someone tell me why the fix for #506 (menu button taking you out to home screen) contained within this uber-big PR?  Feels like they would be pretty separate items and we could merge the menu fix before this much larger change?  Perhaps it's not as simple as I'm making it!

lolkthxbai:
@carloperez if you can't wait I think you can always just clone the fork by @JoeMatt

lolkthxbai:
@Cepheidsong I couldn't get Provenance to recognize the PSX BIOS files. I created a new folder BIOS>com.provenance.psx and put the "scope####.bin" files there.

JoeMatt:
I just finished BIOS MD5 matching so the issues mentioned above should be mostly moot now.
I'm going to expand this to ROM matching as well now that we have so many bin/cue systems.

As far as grainy video, it's whatever the Mednefen core renders. We don't have shaders yet. It's an idea #251 that's been thought about. I don't think anyone with shader experience on iOS has spoken up to implement. I played with some FOSS shaders but nothing was in a format Xcode/iOS understood natively. I understand there's possibly a way to convert between languages/fileformats/versions for shaders. It's all new to me but is something I'd like to implement if i learn how.

lolkthxbai:
@JoeMatt how about the mfi button assignments? its easier to bring up the emulator pause menu and choose start and select as one of the menu options than reassigning them to L2 and R2

QuarterSwede:
@lolthxbai I'd love to see MFI button mapping but having start and select in the pause menu is not a great idea. There are a lot of older games that require select to be used to select options. It would make them hard to play.

JoeMatt:
For the L2/R2 I don't have a good solution. MFi has only so many buttons. More games depend on Start Select than L/R2. Maybe can make it an option to change but that's not high on my list at the moment.

lolkthxbai:
@JoeMatt @QuarterSwede for psx, using select and start hasn't been essential for me. in fact i've only had to use it to start the game or pause the game while L/R2 is used to rotate between weapons or camera controls making gameplay difficult without them. but that's just me, so i get why it was more important the other way around.

last question about buttons, under what circumstances does the emulator pause menu show you P1 Start and P1 Select ?

jasarien:
Mostly only when using the siri remote instead of a real game controller on apple TV.

Cepheidsong:
sega cd bios work now!  Tho the slowdown issue is still weird.  It's not just for psx (slowdown/stuttering) but something is weird about the music for other systems.  The SNES, about 5 seconds in, the music goes into a really slow off-pitch (the game runs normal speed, but choppy).  The Genesis too, but not as bad.  Sega CD and NES seemed to be ok.  I tested with many roms for each to duplicate effect, all same.

Changing the speed option doesn't seem to change anything about the sound choppiness for any of those.

JoeMatt:
@Cepheidsong  Are you possibly building in debug mode? Switch to release configuration if so.

Cepheidsong:
@JoeMatt Yep I am in debug mode!  Good call.  Now I just have to figure out how to get it in release config, it seems to be stuck there for me haha.  But at least that might be my issue.

lolkthxbai:
@Cepheidsong build from xcode again but make sure your target is set to provenance-release or provenancetv-release

mikhail-nci:
For some reason MFI analog sticks are not working, when playing PSX games... I am on the latest release.

lolkthxbai:
same here, playing jersey devil is a real pain without camera controls

mikhail-nci:
After more testing I found that MFI analog sticks tend to work in some games, like NFS: Porsche Unleashed or X-Files, but then D-pads don't seem to do anything there.

lolkthxbai:
Might it have something to do with some games using digital and analog modes for the analog sticks? that's the only reason i could think of

paviro:
Any news on this?

cristofercruz:
I did a bit of testing on a couple of the additions. Here are a few things I found.

Pokemon mini: Rom loading is inconsistent. Most of the time it loads to a gray screen, repeatedly trying to load a ROM will eventually get one to load correctly. As far as playability, I think there are some input methods required by these games which are not present here. Overall seems to just not work that well.

Atari 5200: I am getting a consistent crash when attempting to load a 5200 game after returning to the game library. Steps: 1) Open a 2600 or 5200 game 2) Return to game library 3) Attempt to load a 5200 game; results in crash. On playability, I could be wrong but 5200 seems to be missing MFI mappings to start a game? I had to press the play/pause button on the Apple TV remove to start a game. Once started I could use my MFI controller D-pad and A button.

paviro:
I tried to launch Chrono Chross, Provenance crashes instantly every time.

JoeMatt:
As far as individual ROMs not loading, it's not going to be fixed. It's probably related to the ROM or the emulator core itself and the goal of this project (at least not for me) is not to provide upstream fixes to emulators that are included. Provenance is only a front end, with some platform/compiler specific hacks where necessary.

I'll look into the 5200 thing. Pokemon mini, the roms are super inconsistent all over the net. It's a really odd system and most of the ROMs seem to be user hacks or alpha/beta games. The most successful commercial roms seem to load fine and that's enough for me at least. It is what it is, or not at all. I think the upstream project isn't even under active dev anymore.

I'm mostly working on improving the ROM import and the few bugs listed already. I'll be out on holiday for a couple weeks though so don't expect much traction. I MAY get MD5 ROM import matching before then. I have it semi working.

paviro:
Totally understand :)
Just out of curiosity: It works in open emu is it a different core?

JoeMatt:
I think we do use a different version of Mednafen (the PSX core). Mednafen doesn't put the version number anywhere in their source, that I found, so I kind of forget. Though they're not too far off and Mednafen barely has made any PSX patches in years so... i dunno. I can look at it maybe at some point if it is known to work in OpenEMU but not Prov. Maybe there's some feature I haven't hooked up yet.

Cepheidsong:
I had 2 psx roms to try it out, both get laggy almost right away with the audio like it can't keep up, and sounds off/distorted.  So I went to check libretro which usually has a date on updates on it's cores, says mednafen had an update last month- i tried it out there, but it doesn't even load.  So I checked pcsx-rearmed (interperter) and that had no flaws, ran the same games fine.

This on iphone 6 jailbreak.  In case any of that info helps.

JoeMatt:
Did you try the game speed fix that's been mentioned in this thread already?

Cepheidsong:
Ya, I tried the speed fix on both, and tried all sort of different combos too when that didn't work.  Made sure the build was right and all.  Everything else works flawlessly, even sega cd.  I mention my iphone type because maybe I wondered if it just needed a better processor?  Dunno.

As for operation though, it operates just fine.  Just the sound going sour is all.

JoeMatt:
Perhaps iPhone 6 is too slow. If you're using pcsx-rearmed in a jailbreak release then it uses the dynamic recompiler so it's going to be faster. If it's side loaded then it may still be a faster implementation of PSX. Maybe we'll look into that at some point. After Mednafen though I'm going to be working on Sega Saturn and possibly PSP (some early success here). We'll have to circle back to pcsx-rearmed at a later time.

leoderbois:
Hi,

Work well on iPad Air 2, Apple TV4 and iPhone 6 for me. Tried Tekken 3, Tomb Raider, Tony Hawk Pro Skater, Ape Escape (but miss the analog press on my Nimbus), Ridge Racer, Crash Bandicoot. all works well ! Very nice job 😊

jasarien:
@JoeMatt what is left for this PR? Were you going to handle multi-disc games for PSX?

JoeMatt:
MD5 based disc import and a few user reported bugs with analog stick, and sometimes it's still startinng with the wrong clock timing causing the slow down at start. 
I think maybe still an issue with opening repeatedly. 

I've been super busy, it's summer, out and about, just haven't had the time and when I do it's on paid projects at the moment.

jasarien:
@joematt no worries, wasn't trying to be pushy. 😅 was just wondering what was in progress and whether multi-disc games was on the cards.

JoeMatt:
Multi game should be on the table but honestly I haven't thought about it (was a N64 kid sorry). It's defintently desired but maybe not a blocker for merge. I think just getting the system running correctly consistently and importing non multi tiered roms is 'good enough'. I wouldn't ship that way commercially but I think for our purposes that would be a worth wild win. Honestly I'm somewhat embarrassed it's taking me this long, poor time management on my part I apologize.

jasarien:
My only concern about multi-disc games is Dave sharing. Provenance based it's save file login on open emu, where a single save file is created per ROM. For PSX, this would mean one memory card file per disc, which would mean you wouldn't be able to continue a game when it asks you to insert the next disc (eg, final fantasy). 

With PSX memory card files, I don't see why we couldn't create a global single memory card and use it for all PSX games. This would allow save sharing between discs without having to follow open emu's model of importing multiple discs as a single ROM. 

Please don't be embarrassed or feel pressured to work on this, btw! No one's cracking whips here. 😂

MysticalOS:
Personally, isn't disc combining into one image better solution anyways? most PS1 emulators I used did this. my good ole playstation portable with CFW still uses 1 image files for PS1

XeresRazor:
Mednafen supports playing PBP images as used by the PSP. These support multiple discs per image, and metadata about the game.

Cepheidsong:
Just upgraded to iphone 7 and gave the psx another spin (on the release version).  No jailbreak right now, on 10.3.2.

First thing is everything loads and runs super fast- there's not a single hiccup in any graphics, no lag there.  All other emulators including sega CD have no issues.

But the music in the PSX games still sounds sour.  Like the music is choppy, wavey.  Like the music is lagging even though the graphics aren't.  No drop in frame rate!

I did some research on other people who had similar problems with medafen and psx, and some solutions pointed to needing newer version of medafen, or increasing/doubling the audio buffer size.

imseandavis:
Just wanted to drop in an say great job on everything guys, this is such an awesome product!

brianmcdoogle:
First off, great work to everyone working on this. Yay! I hope multi-disc stuff gets figured out (and maybe in the future cheats).

With iOS 11 and the new Xcode 9 out, I thought I would try to put this on a new iPad. In the cheat.cpp file of PVNES I'm getting this error: "Ordered comparison between pointer and zero ('char *' and 'int')" and it is being triggered at line 219 "while(fgets(linebuf,2048,fp)>0)". Don't know if it is an issue with Xcode 9 and the new iOS 11 but thought I'd share.

jasarien:
@brianmcdoogle Yeah, the compiler in Xcode 9 is a bit more fussy than previous versions. This is fixed in the master branch, but hasn’t been merged into this one yet. You can apply a hot fix yourself, by changing: `while(fgets(linebuf,2048,fp)>0)` to `while(fgets(linebuf,2048,fp)>(void *)0)`

armaanaki:
Unsure why, but when I attempt to compile this I get a Semantic Issue, namely "Call to 'abs' is ambiguous". Any solution?

jasarien:
@armaanaki yeah, this branch is a few commits behind master and hasn’t been updated for Xcode 9 or iOS 11. It will need to merge in the edits to placate the more fussy compilers in Xcode 9.

lanSmith:
I tried the latest build of this tonight and saw some audio and video issues. I tried three PSX games (Gran Turismo 2, MGS2, and PaRappa) all of them had really low frame rate (20-30 on menus & ~15 in game) and choppy audio. I tried the game speed reset mentioned above but all three settings behave the same way. The md5s of my scph files match the recommendations on the Mednafen site so I don't think it is an engine issue.

armaanaki:
Just tested new changes, working on iPhone 6s Plus iOS 11.0.2

sledgehammer617:
Any possible fixes to the distorted audio in PSX games?

Cepheidsong:
On similar emulator I've been able to solve similar distorted audio by raising buffer or using a CD Image Cache option.  Maybe in the build itself that can be adjusted to be bigger or on.

JoeMatt:
I've pushed a number of updated recently. I believe this PR to be release ready for all intents and purposes. There may be some outlying non-critical bugs. It could use some community testing though for those that are tracking this PR.

The major features / bug fixes

- Import ROMs by MD5 or filename when extension has multiple destinations
- BIOS recognition and copy to proper destination by MD5 or filename
- Mednafen occasional slow playback fixed (the previous work around was selecting Normal playback speed. Any audio stuttering that isn't fixed by that work around is either too slow device or user is running Debug configuration. Please report issues with playback that are still fixed by workaround since that implies my playback calculation fix isn't correct)
- Fix crash in Atari 800 core when loading a second ROM

I've been playing around with this branch and everything seems ok to me. There are some issues to fix that aren't related to this branch

- Some games have poor button mapping - would need to add ability to remap #56 
- Some games don't respond to joystick properly - need a system to customise controller emulation type.
- Pixulated rendering. There's a pull request to add CRT shaders. It runs too slow on PSX for Apple TV 4 at the moment #608 in my testing but works well for other systems. Either #608 needs to be optimised or 4K Apple TV with updated GPU is required. Currently don't have a way to notify users of degraded performance based on hardware / settings.
- Multi-disc mode. I haven't tested this at all. Was a concern brought up by @jasarien . Possibly this can be handled after merge in a separate bug ticket if it's not considered a blocker for merge.

Hopefully we can get this approved and merged soon. XMAS preset for the community!

braindx:
I can take a pass at optimizing the shader from #608. The expensive bits are the multiple texture samples for the blurring so maybe I’ll refactor that to a separable blur. Could also just lower the # of samples when running on Apple TV 4 through a preprocessor macro. I don’t have my Mac with me right now so I won’t be able to get to it for a little bit. I also don’t have an Apple TV 4 so someone else will have to test it out and let me know how the performance is.

JoeMatt:
@braindx That's very generous thank you. The shader looks amazing on all the other cores. Looks great on PSX just lowers the frame rate a bit too much. 

I have a branch here where I've merged #608 plus this branch and a few others together to test in combination.
https://github.com/JoeMatt/Provenance/tree/Mednafen+Atari800+Poke+OthersUpdates

JoeMatt:
As far as the preprocessor macro, I estimate the power of an ATV is around an iPad 4 or iPhone 6. I believe it's a slightly under-clocked A8. Perhaps 2 versions of the shader so the user can select quality vs performance. Ideally we'd have a way to rate the processing requirements of the core + hardware running on and smartly select the optimal version. It would just require some man power on coming up with rating estimations for each device and core combination. For now a manual low / high setting would be adaquette. Perhaps we could make it changeable  during run time like OpenEMU has for shaders. Now that we have the architecture it shouldn't be difficult to copy other FOSS shaders into the project. I realise iOS has a somewhat unstandardised shader language compared to the other cross platform emulators. I have basically no knowledge and hit a brick wall trying to get shaders from other projects to work. Now that we have a working base from you though it gives a framework to test other shaders wether copied or cross converted from other shader standards. Once we have a few options an at runtime menu to change shaders would be preferable.

appletv4:
For neogeo 

I need to import the neogeo.zip into provenance?
And what version of neogeo rom set do I need to use

braindx:
Just getting a chance to take another look at this now... forget everything I said about a separable blur... I forgot I ended up only using a 1D blur and taking 3 taps so there's nothing to be done there. That also means it _should_ be relatively cheap.

The CRT shader is performed on the GPU as a pixel shader and so conceptually, since all the emulation cores are run exclusively on the CPU (as far as I'm aware), the emulation should be equally performant on all emulator cores since the GPU _should_ always have the full frame to execute that pixel shader. In practice, however, at the end of each frame that the emulator core finishes I believe the renderer is blocking the emulator core from beginning its work on the next frame until the frame is rendered due to a @synchronized block in PVGLViewController. From a cursory glance, it seems this is probably there because the framebuffer generated by the emulator core needs to be transferred to the GPU and you don't want the emulator core overwriting it as it works on the next frame before the GPU has gotten a hold of it. The better approach would be to make the framebuffers double-buffered so that the emulator core alternates which framebuffer it's rendering to and the GPU can operate on one while the emulator core writes to the other.

I'll mess around with a couple of things on my end here but if you're available to test some changes out on your Apple TV 4 let me know!

JoeMatt:
Yeah that sounds amazing if you can improve the general rendering infrastructure, it would give us a lot of room to breathe on implementing more advance cores. We copied the rendering structure from OpenEMU so i assumed it was all up to best practices but if there's improvements to be made, especially since we're targeting much more restricted hardware than OpenEMU, it would open us up to implementing more advanced gen cores. Would love to test anything you put out.

braindx:
The general rendering infrastructure seems fine. The core and renderer are already on separate threads. The only issue I could see was that the mutex for the emulator core remained locked while the framebuffer was being uploaded to the GPU which may or may not have been impacting performance particularly on cores with larger internal resolutions. Check out https://github.com/braindx/Provenance/tree/doublebuffered and let me know if that helps or hurts PSX performance on Apple TV 4. What are the steps for running PSX games by the way? What files need to be uploaded and in what formats?

On another note, could you explain the widescreen flag that is now used in the renderer? I'm not sure why that's necessary and/or what it's supposed to be doing.

braindx:
Don't bother testing that. I've confirmed locally that the framebuffer copy is too expensive to be worth doing. I'm going to add support to PVEmulatorCore for double-buffered framebuffers and will implement that for the PSX core for now. This will assume that the emulator core keeps two framebuffers around and writes to one while the other is returned from [self.emulatorCore videoBuffer]. For cores where this isn't implemented, it will operate as before.

I've also found the wiki for getting PSX games up and running, so I've got that working now so I can test locally. As far as the widescreen flag, it seems like that's just a hack to address the fact that the frame isn't written at the top-left (0,0) of the framebuffer for PSX, a situation which isn't currently handled correctly by the renderer. Is that right or is there something else going on? If that's all it is, I can fix that up as well.

XxXStefannXxX:
I know I am not in the right discussion room but i have a request to make, to all programmers working on Provenance. Would it be possible to make r2 and l2 buttons in the app? Some update or smth? I’m not an expert or anything, and don’t know how its made but it would be great to have all buttons that were on Ps1 controller. Thanks guys

JoeMatt:
@XxXStefannXxX You know one comment is enough. We get alerts. There are 3 part time people working on about 12 other things right now. Until someone else steps up and fixes #644 you'll have to wait, and please search next and and don't double post or post in closed threads.

XxXStefannXxX:
Ok, sorry about that. After I posted in one discussion I saw that it was closed and relinked to the other one so I reposted. Thank you for your reply, and would much apreciate it if you could make the buttons. Thank you and sorry again.
________________________________

XxXStefannXxX:
I poster there also, didn’t know you guys have seen my post already. Sorry again.
________________________________

PR #582, opened 2017-07-03
8bitdo Zero Controller Support

justinweiss:
The 8bitdo Zero has its buttons reversed from other 8bitdo controllers -- On a SNES game, the bottom button acts like 'A', not 'B', and the top button acts like 'Y', not 'X'. This patch adds a new controller class for the 8bitdo Zero that maps the controller buttons to the correct inputs.

### DIFF
--- a/Provenance/Controller/PVControllerManager.m
+++ b/Provenance/Controller/PVControllerManager.m
@@ -11,6 +11,7 @@
 #import "PViCadeController.h"
 #import "kICadeControllerSetting.h"
 #import "PViCade8BitdoController.h"
+#import "PViCade8BitdoZeroController.h"
 #import "PViCadeController.h"
 
 NSString * const PVControllerManagerControllerReassignedNotification = @"PVControllerManagerControllerReassignedNotification";
@@ -111,7 +112,7 @@ NSString * const PVControllerManagerControllerReassignedNotification = @"PVContr
     }
     
     BOOL assigned = NO;
-    if ([controller isKindOfClass:[PViCade8BitdoController class]]) {
+    if ([controller isKindOfClass:[PViCade8BitdoController class]] || [controller isKindOfClass:[PViCade8BitdoZeroController class]]) {
         // For 8Bitdo, we set to listen again for controllers after disconnecting
         // so we can detect when they connect again
         if (self.iCadeController) {
--- a/Provenance/Emulator/PVEmulatorViewController.m
+++ b/Provenance/Emulator/PVEmulatorViewController.m
@@ -22,6 +22,7 @@
 #import "PVEmulatorConfiguration.h"
 #import "PVControllerManager.h"
 #import "PViCade8BitdoController.h"
+#import "PViCade8BitdoZeroController.h"
 
 @interface PVEmulatorViewController ()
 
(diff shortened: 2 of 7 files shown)


Provenance/Emulator/PVEmulatorViewController.m
```diff
@@ -841,7 +842,7 @@ - (void)controllerDidConnect:(NSNotification *)note
 {
     GCController *controller = [note object];
     // 8Bitdo controllers don't have a pause button, so don't hide the menu
-    if (![controller isKindOfClass:[PViCade8BitdoController class]]) {
+    if (![controller isKindOfClass:[PViCade8BitdoController class]] || ![controller isKindOfClass:[PViCade8BitdoZeroController class]]) {
```
justinweiss: At some point, it might make sense to have the controller tell this class whether it has a pause button, but with only two examples, I wasn't ready to make that change.

Provenance/Controller/PVControllerManager.m
```diff
@@ -111,7 +112,7 @@ - (void)handleControllerDidDisconnect:(NSNotification *)note
     }
     
     BOOL assigned = NO;
-    if ([controller isKindOfClass:[PViCade8BitdoController class]]) {
+    if ([controller isKindOfClass:[PViCade8BitdoController class]] || [controller isKindOfClass:[PViCade8BitdoZeroController class]]) {
```
justinweiss: Same thing as the "menu button" -- this could be a controlled by a property of the 8bitdocontroller classes, but I wasn't quite ready to extract it yet.

bakkiejochurt:
Would you mind building an IPA with this support, because I have an 8bitdo zero, but I don’t have a Mac and my laptop isn’t good enough to run a virtual one(so I can’t get Xcode).

Devedander:
Seconding desire for an Zero compatible build!

Davu67:
I have the same problem and i customized the firmware of my 8bitdo zero, just visit the oficial site! They have all the tools to do that. I switch the buttons directly jn the firmware and all works great!

Devedander:
Davu67 can you give a little more detail?  I went to 8bitdo website but didn't see any firmware editing tools... I did find remapping tools but they don't have one for ios...

Davu67:
It seems they delete the page with the manual and the tool to do that. I will search in my pc a little more to you. I wiil help you relax. I done this two times because i have two 8bitzero controller. Wait for it...

Davu67:
after some hours searching I founded! you will need a windows pc to do this. the firmware is here [http://download.8bitdo.com/Firmware/GamePad/bluetooth_firmware_V2.70.zip](url) inside have a folder named "remap tools" se the instructions inside. I did it the remapping of my two 8bitdo zero with  this. If you have any problem let me know, maybe I can help you,

JoeMatt:
I added this in commit 8ab62bb

Don't have one to test though so someone should test that. One is on the way though from Amazon in a few days.

Grandtheft430:
@Davu67 i have the same problem as the guy in #650 has - i've launched the remap software, but it says no controller. In windows' device list it doesn't show up as well - it says there's an unknown device. How do i plug it in to remap the buttons?

PR #594, opened 2017-09-14
Fixed background gradient on various pieces of art

willco007:
Some of the icon background gradients end before they get to the edge of the artwork which causes the masking on tvOS icons to not look correct. I extended the gradient to the edges.

Sorry for the merge commits as well as the artwork, still getting the hang of this github thing. :)


sevdestruct:
@willco007: could you by chance post a screenshot of the error and the fix?

willco007:
Turns out, this fix was already in a prior pull request, commit 5bbcf539, I had submitted a while back. I didn't think I had pushed it but I guess I did. :) Closing.

PR #604, opened 2017-10-28
Speed up app launch (specially on large libraries)

leolobato:
Only trigger a library refresh if a rom lookup was triggered.

I've added 1888 nes/snes games to Provenance on an iPhone 6 and it would take 38 seconds to be responsive on launch. Now it's immediate.

I'm not sure why we run the game importer for each ROM that is already on the library, but the finish handler was called for each of those, which would trigger a library refresh (which can block the main thread for almost a second). I've modified the code to only refresh the library if the importer actually fetched new data for that ROM.

Library refresh speed can still be improved, but less than a second sounds OK to me right now.


appletv4:
Is this also for Apple TV ?

leolobato:
I assume so, but I didn't test on it.

Issue #54, opened 2015-02-11
Smoothing & scan line filters


cplr:
Definitely not a high priority item, but I think generally people would love different smoothing and "scan line" filters. Even better would be curved-CRT shaders :). This project has some potential shaders that could be used for the curved-CRT look: https://github.com/Swordfish90/cool-retro-term

latzen79:
Oh!!! I need scanlines for my appletv , scanlines and moré filters

Naatan:
This would be amazing! +1

MysticalOS:
http://filthypants.blogspot.com/2012/03/xbr-vs-hqx-interpolation-filter.html
That's the one I'd love to see not sure of performance impact though. however ios devices are a lot more powerful than they used to be, you could possibly add such a filter with device restriction or at very least warning that they probably shouldn't be enabled on weaker devices.

MysticalOS:
iNDS recently added a bunch of filters and they are pretty great. Worth checking out. some at higher performance cost than others but modern iOS devices are pretty strong. i tried most of those filters with no performance loss, and that's with an emulator that's already inherently more demanding.
https://github.com/WilliamLCobb/iNDS

dasilvetz:
Having seen the various filters on the NES Classic console, have to say, would be great if Provenance supported similar features! +1 [http://www.theverge.com/2016/9/30/13089242/nes-classic-nintendo-new-features-multiple-display-modes](url)

gklka:
https://github.com/OpenEmu/OpenEmu – OpenEmu for Mac also has a lof of excellent filters, maybe they can be used in this project too.

circa-one:
Surprised this hasn't been implemented. Especially for a tvOS app. Would love to see it happen!

braindx:
I've just added this, along with enabling 4K resolution for Apple TV 4Ks, to my branch here:
https://github.com/braindx/Provenance/tree/crt-filter

The shader is one that I just wrote but I am releasing it into the public domain. It's heavily inspired by Timothy Lottes' shader, but that one is expensive to perform in a single pass and doesn't scale down well. I wanted something that was going to be performant on my Apple TV 4K and still look decent on my phone. Of course, now that it's hooked up, the shader can be replaced with whatever shader you want by replacing the contents of shader_crt_fragment.glsl. There are also a number of settings that can be tweaked at the top of the file.

Let me know if you'd like to make this a formal pull request.

gklka:
[...]

PR #608, opened 2017-11-09
Added CRT shader and enabled 4K resolution for Apple TV 4K.

braindx:
The CRT shader used is one that I just wrote but I am releasing it into the public domain. It's heavily inspired by Timothy Lottes' shader, but that one is expensive to perform in a single pass and doesn't scale down well. I wanted something that was going to be performant on my Apple TV 4K and still look decent on my phone. Of course, now that it's hooked up, the shader can be replaced with whatever shader you want by replacing the contents of shader_crt_fragment.glsl. There are also a number of settings that can be tweaked at the top of the file.

CRT Filter can be enabled/disabled from the settings menu. It's off by default.

4K was enabled simply by adding a 2x launch image, not a big deal.

This pull request satisfies issue #54. Ideally, one could select from a list of shaders rather than just a single toggle, but it's a start.


appletv4:
Been waiting 2 years for this , thanks 
@jasarien release the kraken

appletv4:
Testing this now 
Btw they changed the way you choose a branch in Xcode 9 very confusing 

The crt filter looks and works great , is better with it on

DanFessler:
just tested this locally myself and it runs and looks great.  Definitely think this should be merged in

JoeMatt:
hm, I merged this branch into a testing branch on my fork and now it's showing my forks commits on this PR for some reason. I've done this method before, not sure what went wrong here.

JoeMatt:
Did the same steps on #604 into my testing branch and that PR isn't showing my merge history. Not sure what happened here.

jasarien:
@joe I’m a bit confused as to whether or not the mednafen PR included the CRT shader code or not.

PR #612, opened 2017-11-17
Added ability to unzip rar and 7z formats.

tkunstek:
Added ability to unzip rar and 7z formats. Now also finds files in subdirectories of all archive types.


JoeMatt:
In #566 I already added 7z through LzmaSDKObjC framework. It is missing RAR. I'd merge but there's a bunch of conflicts since I already added code paths for 7z.

jasarien:
@tkunstek would it be possible to remove the 7z support from this PR so that it only includes RAR? Or are the two formats provided by the same lib?

tkunstek:
Library used supports both formats. You can reject this PR and I'll come back through and add 7z separately.

jasarien:
Just to clarify it’s the RAR support that needs adding separately - 7z was added with @Joe’a megaPR

tkunstek:
Ah. I can still add RAR too if it will benefit someone.

JoeMatt:
If this were to be added it would make sense to remove my LzmaSDKObjC dependency and depend on SARUnArchiveANY solely to reduce dependency management.

PR #613, opened 2017-11-19
added iPhone 8, iPhone 8Plus and iPhone X support

sonique6784:
- iPhone 8, 8 Plus and X support Taptic Engine
 - iPhone X moved SELECT and START button to prevent overlapping with new
   Home "button" bar at the bottom of the screen
- iPhone X moved MENU button to bottom left on iPhone X

### DIFF
--- a/Provenance/Controller/PVControllerViewController.m
+++ b/Provenance/Controller/PVControllerViewController.m
@@ -235,7 +235,7 @@
 		}
 		else if ([controlType isEqualToString:PVStartButton])
 		{
-            CGFloat yPadding = ([UIDevice isIphoneX] ? 50 : 10);
+			CGFloat yPadding = ([UIDevice isIphoneX] ? 50 : 10);
 			CGRect startFrame = CGRectMake((self.view.frame.size.width - controlSize.width) / 2, self.view.frame.size.height - controlSize.height - yPadding, controlSize.width, controlSize.height);
 			
 			if (!self.startButton)
@@ -257,7 +257,7 @@
 		}
 		else if ([controlType isEqualToString:PVSelectButton])
 		{
-            CGFloat yPadding = ([UIDevice isIphoneX] ? 30 : 10);
+			CGFloat yPadding = ([UIDevice isIphoneX] ? 30 : 10);
 			CGRect selectFrame = CGRectMake((self.view.frame.size.width - controlSize.width) / 2, self.view.frame.size.height - (controlSize.height * 2) - (yPadding * 2), controlSize.width, controlSize.height);
 			
 			if (!self.selectButton)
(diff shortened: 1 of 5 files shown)


Provenance/Controller/PVControllerViewController.m
```diff
@@ -235,7 +235,7 @@ - (void)setupTouchControls
 		}
 		else if ([controlType isEqualToString:PVStartButton])
 		{
-			CGFloat yPadding = 10;
+            CGFloat yPadding = ([UIDevice isIphoneX] ? 50 : 10);
```
DwayneCoussement: Use `view.safeAreaInsets.bottom` to determine the frame. Checking for iPhoneX is not the right way to do this.

gklka:
Works fine for me.

Provenance/Emulator/PVEmulatorViewController.m
```diff
@@ -185,7 +185,7 @@ - (void)viewDidLoad
 	
 	CGFloat alpha = [[PVSettingsModel sharedInstance] controllerOpacity];
 	self.menuButton = [UIButton buttonWithType:UIButtonTypeCustom];
-    [self.menuButton setFrame:CGRectMake( ([UIDevice isIphoneX] ? 10 : ([[self view] bounds].size.width - 62) / 2 ), ([UIDevice isIphoneX] ? 740 : 10), 62, 22)];
+    [self.menuButton setFrame:CGRectMake( ([UIDevice isIphoneX] ? 10 : ([[self view] bounds].size.width - 62) / 2 ), ([UIDevice isIphoneX] ? [[self view] bounds].size.height - 60 : 10), 62, 22)];
```
DwayneCoussement: Again, calculate with `safeArea`, don't use a device check here. otherwise next year you might end up having 3 devices in there.

jasarien:
The way this has been done may work for now, but it’s not very maintainable, as @DwayneCoussement has mentioned. Before you know it, after Apple releases the iPhone X Plus or whatever, we’ll be needing device checks all over the place.

It’s more work, but the proper way to acheive this is to re-do the UI layer with modern techniques, such as autolayout and safe area insets.

gklka:
I agree, UI needs to be done with AutoLayout and the modern stuff to support the future generation of iOS devices. But until then: menu button is not visible and Start is barely touchable on iPhone X, Start button is covered by the Dock gesture on iPad... So I think this could act as a hotfix.

DwayneCoussement:
You could still use the frame layouts, and get the layoutFrame from the safeArea. Which could be the hotfix, without introducing AutoLayout all over the place. The advantage would be that it'd be as maintainable as it is right now.

jasarien:
@sonique6784 do you think you could implement this using the safe area insets instead of querying for different device models? You'll also need to resolve some conflicts since other PRs have been merged since this was raised.

jasarien:
I’m closing this PR in favour of #629, which uses the safe area insets.

Issue #601, opened 2017-10-09
iCade 8Bitdo FC30 Controller : Bad mapping


Canadadry:
Hi I've tried to play with a 8Bitdo FC30 Controller on my iPhone IOS10 (French). The dpad is not mapped correctly I cant move in any direction but other button are working fine. 

So I dig up a little and I found my controller is not sending the same key as defined in iCadeReaderView.h

```
/*
  What I should have => what I really have
 UP ON,OFF  = w,e    => z,e
 RT ON,OFF  = d,c    => d,c
 DN ON,OFF  = x,z    => x,w
 LT ON,OFF  = a,q    => q,a
 A  ON,OFF  = y,t    => y,t (Y)
 B  ON,OFF  = h,r    => h,r (B)
 C  ON,OFF  = u,f    => u,f (A)
 D  ON,OFF  = j,n    => j,n (X)
 E  ON,OFF  = i,m    => i,[,] (R)
 F  ON,OFF  = k,p    => k,p (L)
 G  ON,OFF  = o,g    => o,g (Select)
 H  ON,OFF  = l,v    => l,v (Start)
*/
```

It seems that the controller is sending keychar as a AZERTY keyboard (since my iphone is in french) and provenance expect the keychar form an QWERTY keyboard. 

I guess we should extract the mapping as localisable data. 

For now I have edited iCadeRedearView.m and it work fine, but that cannot be a patch.  Here what I have changed to make it work. 

```
//static const char *ON_STATES  = "wdxayhujikol"; //wdxazhujikol for German keyboard layout
//static const char *OFF_STATES = "eczqtrfnmpgv"; //ecyqtrfnmpgv for German keyboard layout

static const char *ON_STATES  = "zdxqyhujikol"; //wdxazhujikol for French keyboard layout
static const char *OFF_STATES = "ecwatrfn,pgv"; //ecyqtrfnmpgv for French keyboard layout
```

Vahron:
Hi,

Just to confirm I have the exact same issue with a SFC30 controller (it's the same as the FC30 and has the same mapping, only its form factor is different) on a French iPhone 7 (iOS 11). Your workaround will help while waiting for a proper fix, thank you!

CodyJLeeApps:
Thinking about this outside the context of this app:
Based on my understanding of the 8Bitdo Controller (and the company instructions) the keyboard you have activated in the device needs to be “English (US)” for the controller to work properly.

[...]

PR #614, opened 2017-11-22
Add iCade controller support for FR and DE on tvOS

viteinfinite:


viteinfinite:
This also provides a proper fix for #601 when playing on an Apple TV.
Tested in French with my 8bitdo FC30 zero.

viteinfinite:
🎉

Provenance/Controller/iCade/iCadeReaderView.m
```diff
@@ -115,13 +126,32 @@ - (void)deleteBackward {
 
 #pragma mark - keys
 
+- (void)applyKeyMappingForCurrentLocale {
+#if TARGET_OS_TV
+    NSString *localeIdentifier = [[NSLocale currentLocale] localeIdentifier];
```
jasarien: Is there any reason why this is only done on Apple TV and not iOS?

Canadadry:
@viteinfinite sorry, this is not working for other plateforme than AppleTv. Why did you needed to put de the `#if`statement ? 

If Have removed it and the code is still not working on my iPhone. My local identifier is `en_FR`

Issue #522, opened 2017-01-28
B Button Triggers App Menu on SteelSeries Nimbus with AppleTV 4


bastardsby:
Hi, first of all, thank you so much for all the work you've put into this beautiful app. I've been trying it out for the last couple of days and had a lot of fun playing old NES and Genesis games again.

I'm not sure if this is a bug with the emulator or something else, but the SteelSeries Nimbus' B button seems to trigger both the app menu (not the ROM's menu) along with the expected in-game B button behavior. For example, when playing Super Bomberman 3 for SNES, if I press the B button to set a bomb, the bomb will be set, but the app menu will also be brought up asking if I want to Return to Game Library or Resume. The Menu button on the Nimbus has the same functionality as the B button (which is also the case when navigating the AppleTV interface outside the Provenance app).

I'm using Provenance 1.3.2 on an AppleTV 4 running tvOS 10.1.1. This is my first time posting anything on GitHub, so apologies if the answer to this issue is obvious or if I left something out.

zfisher323:
I can confirm the exact same behavior on my steel series nimbus running 1.3.2.

appletv4:
Remove the app from the background and reopen

TomR01:
Yeah I have this problem too - is there a fix?

bastardsby:
@appletv4 I tried removing the app and reopening, but no luck.

valdetero:
I just installed it on my Apple TV and noticed the same issue.

lolkthxbai:
tvOS 10.1.1 Provenance 1.3.2 rev ec652bb
Nimbus Steelseries Ver. 1.6.3.1
I can't reproduce the issue. I added Super Bomberman 3 to my library and was able to place bombs (B Button) without the in app menu being triggered.
Update Provenance if you can and also update your controller using the Nimbus app from the AppStore. Try it again and post back your results.

valdetero:
@lolkthxbai I got latest right before I installed it 2 days ago. I'll see if the controller has an update.

valdetero:
I'm running tvOS 10.1.1, Nimbus 1.6.4.1, Provence 1.3.2 (659), Revision bc6a42a and it is a problem on Bomberman 3. I changed the iCade controller types but it didn't seem to make a difference. Should the Nimbus be Standard or SteelSeries Free?
Are you using older versions than me?

lolkthxbai:
[...]

PR #628, opened 2017-12-22
Fix for iCade + MFi Controller Conflict (tvOS)

DanielFontes:
On the Apple TV the Menu or Home button didn't work as expected while in emulation mode. Instead of showing the pause menu the app jumped into background. In order to change this UITapGestureRecognizer hast to be listening on the keyWindow of the App instead of the current view. The UITapGestureRecognizer should also be removed in dealloc.

### DIFF
--- a/Provenance/Emulator/PVEmulatorViewController.m
+++ b/Provenance/Emulator/PVEmulatorViewController.m
@@ -72,6 +72,11 @@ void uncaughtExceptionHandler(NSException *exception)
 
 - (void)dealloc
 {
+    if (self.menuGestureRecognizer) {
+        [[UIApplication sharedApplication].keyWindow removeGestureRecognizer:self.menuGestureRecognizer];
+        [self setMenuGestureRecognizer:nil];
+    }
+    
     [self.emulatorCore stopEmulation]; //Leave emulation loop first
     [self.gameAudio stopAudio];
 
@@ -334,11 +339,11 @@ void uncaughtExceptionHandler(NSException *exception)
 #if TARGET_OS_TV
     // Adding a tap gesture recognizer for the menu type will override the default 'back' functionality of tvOS
     if (!_menuGestureRecognizer) {
-        _menuGestureRecognizer = [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(controllerPauseButtonPressed:)];
-        _menuGestureRecognizer.allowedPressTypes = @[@(UIPressTypeMenu)];
+        self.menuGestureRecognizer = [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(controllerPauseButtonPressed:)];
+        self.menuGestureRecognizer.allowedPressTypes = @[@(UIPressTypeMenu)];
+        
+        [[UIApplication sharedApplication].keyWindow addGestureRecognizer:self.menuGestureRecognizer];
     }
-    
-    [self.view addGestureRecognizer:_menuGestureRecognizer];
 #else
 	__weak PVEmulatorViewController *weakSelf = self;
 	for (GCController *controller in [GCController controllers])


jasarien:
It’s been working fine for me with the Nimbus and the Stratus XL without this patch.

I wonder why...?

DanielFontes:
So while running a game (I've only tested SNES games) you get the pause menu when you press the "Menu" button on your Apple remote? For me it is only working like that with my patch. Otherwise Provenance is going into the background.

jasarien:
Yep, Apple TV remote menu button and MFi controller pause buttons both bring up the menu. 

http://d.pr/v/wAKN8L

jasarien:
You using the latest code from master? Version 1.4?

DanielFontes:
Really interesting because it's definitely not the same behavior on my Apple TV with the latest tvOS. I wonder why...

DanielFontes:
Yep tested latest release version as well as running the latest commit. Both the same.

DanielFontes:
I have to test it more in the next days...

jasarien:
Do you have an MFi controller? Does it behave the same?

jasarien:
I'm testing your PR to see if has any negative affect in my setup. If not, I guess I'll just merge it. If it fixes it for you, and doesn't break it for me, that's a win-win, right?

DanielFontes:
No I only have a 8bitdo SFC30 controller with no home / menu button.

jasarien:
So, the bad news is that your fix has the opposite affect in my setup. It makes the menu button quit the game and return to the game library. I reverted back to master and the menu button works again in my setup.

There is no good news. 😅

DanielFontes:
😂 ok but now I really want to know why...

jasarien:
@kwant mentioned over on #522 that this occurs when both an iCade controller and MFi controller are connected at the same time. Is this the case in your situation @DanielFontes? Could you try with only one MFi connected or the iCade on it's own?

This is starting to make sense as to why I don't see this issue as I don't have an iCade controller. People with iCade controllers probably see it a lot as they have the iCade and the Apple TV remote connected at the same time, and the Apple TV remote counts as an MFi controller.

jasarien:
I asked about this issue on Twitter and a helpful person mentioned that it is probably something interfering with the Focus Environment on tvOS. I think the iCade Reader View might be stealing focus (as it would need to do to intercept the key command sent by the iCade controller), so the pause handler in the emulator view controller doesn't receive the button press event (or the tap gesture recogniser event), because it's probably swallowed by by the iCade reader view. If we can confirm that this only happens with an iCade controller and MFi controller are connected, then we can start to narrow this down.

Kwant:
After some testing I found the following on my setup: 
When only the mfi controller is connected, all buttons are working properly. Problems occur as soon as the iCade controller setting is enabled. Both the "menu" and "B" buttons wil trigger a return to the home screen without showing the pause menu. Disabling iCade does not fix the issue. After closing provenance from the app switcher, the behaviour returns to normal.

I'm using a SteelSeries nimbus and a 8bitdo nes30 pro. The 8bitdo does not have to be connected to reproduce the issue on my end.

DanielFontes:
@jasarien it's exactly as you guessed. It seems the iCadeReaderView is stealing the focus, that's why the UITapGestureRecognizer stops working on the view of the PVEmulatorViewController.

The home button is working fine without my "fix" when you do not use an iCade controller. As soon as you connect an iCade controller it breaks.

sevdestruct:
To confirm @Kwant's findings, I also have MFi SteelSeries Nimbus and some 8bitdo controllers. Can't use them simultaneously either. Only way it works is just one 8bitdo or two Nimbus', but never 8bitdo + Nimbus. Turning on iCade support renders the Nimbus useless as B button and Menu becomes Home and when that 8bitdo iCade is selected in iCade Controller settings.

@DanielFontes : any progress on this?

vincentsaluzzo:
I've trying the fix of @DanielFontes in my use case. Because I just use a iCade controller (8Bitdo NES30 pro) and in my case, when I just pause game with controllers, the menu doesn't appear, so I use the Apple TV remote to do, but the menu item directly goes to my Apple TV home. By using this branch, I'm able to display game menu directly by taping "menu" button on my remote

Maybe, to ensure there is no problem with standard MFi controller, maybe we can simply determine the view to add gesture recognizer by checking if there only iCade controller binded ? I will try

vincentsaluzzo:
I'm trying with something like this : 

```objc
    PVControllerManager* controllerManager = [PVControllerManager sharedManager];
    if (controllerManager.iCadeController && ([controllerManager.player1 isEqual:controllerManager.iCadeController] || ([controllerManager.player1 isEqual:controllerManager.iCadeController]))) {
        [[UIApplication sharedApplication].keyWindow addGestureRecognizer:_menuGestureRecognizer];
    } else {
        [self.view addGestureRecognizer:_menuGestureRecognizer];
    }
```

But that's ok only if we have defined player 1 or player 2 BEFORE playing. It seems there is a auto discovering if iCade was used when playing to switch automatically to this mode because even if I haven't select it, my 8bitdo are recognized... I need to check more.

jasarien:
The right-hand-side of the `||` should refer to player 2, not player 1, right?

vincentsaluzzo:
Yes you're right ! I've test with the correct code also and the problem also here, I need to go deep in Controller management to understand how it works.

JoeMatt:
If the iCade Reader View is stealing GCController events, the solution is probably to just have it resign active when the menu is up, or conversely, have it only active when the PVEmulatorViewController is the front most view, enabling and disabling during segue's

JoeMatt:
I merged this but I'm going to revert.

iCade is totally dead on tvOS now. Apple changed the way keyboard pairing works, it generates a random number sequence to type in and the iCade controllers can't match them.

Also in my testing without an iCade in, home button kills the game

PR #629, opened 2017-12-22
Feature/iphone x controls layout

SteveBarnegren:
Fixes some layout issues on iPhone X.

**Landscape before / after:**

- Notch was overlapping either the d-pad and the button group, depending on orientation
- Start button overlapping home indicator

landscape

**Portrait before / after:**

- Menu button was inaccessible as it was behind the notch
- Shoulder buttons high up in the status bar
- Start button overlapping home indicator

portrait

**Home Indicator**

I've also made some changes to how the home indicator is handled. There's basically 3 options:

- Have it always visible, and bright white, which is distracting.
- Use `prefersHomeIndicatorAutoHidden`, which hides it. Unfortunately it shows it again every time a button is pressed, which is more distracting.
- Use `preferredScreenEdgesDeferringSystemGestures` which shows it all the time, but makes it translucent (also requires two swipes to go home).

I thought we might be able to combine the two options to hide it all the time, but it's not possible, so I've gone for the last option as it's the most inconspicuous.

Note that the off-centre emulator in the screenshots isn't related to any changes in this PR, it seems to be an issue with the Playstation emu.

I've only tried Crash Bandicoot so far, but it runs perfectly on the iPhone X!

### DIFF
--- a/Provenance/Controller/PVControllerViewController.m
+++ b/Provenance/Controller/PVControllerViewController.m
@@ -119,6 +119,12 @@
 - (void)setupTouchControls
 {
 #if !TARGET_OS_TV
+    
+    UIEdgeInsets safeAreaInsets = UIEdgeInsetsZero;
+    if (@available(iOS 11.0, *)) {
+        safeAreaInsets = self.view.safeAreaInsets;
+    }
+    
 	CGFloat alpha = [[PVSettingsModel sharedInstance] controllerOpacity];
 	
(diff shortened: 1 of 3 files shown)


jasarien:
There must be a way to hide the home indicator even when the user interacts with the screen. Other games do it. 🤔

jasarien:
Ah it seems they use the deferred option. That’s cool with me.

Issue #646, opened 2018-01-10
iPhone X Home Bar/Indicator doesn't fade away in play after touch input idle.


sevdestruct:
[Got an iPhone X last night!]
Noticed the new _Home Bar on iPhone X isn't fading away once in game play_ as it does in video players such as Plex, in which it returns with a tap, as doe the video player interace, and then fades away again after a beat, once the video player controls recede. I'd like to suggest the preferred behavior: **to fade off after touch input idle**.

This was more noticeable when using a gamepad (such as [Game Vice](https://www.amazon.com/iPhone-Controller-Gamepad-Gamevice-Certified/dp/B0786F2VV3)) and detecting this, Provenance hides the entire overlay interface, except for this. Just stays on over your game the whole time.

sevdestruct:
I caught that some iPhone X layout issues were resolved in #629, now closed—including the opacity of the home bar—but not this particular detail.a

sevdestruct:
Submitted fix for this in PR #647 - Working great for me. Any others wanna give it a go on their iPhone X?

PR #647, opened 2018-01-10
Fix for #646: Hide Home Indicator when gamepad/controller is connected (iPhone X) [Incomplete]

sevdestruct:
Attempt to make Home Indicator hide when a controller is connected and overlay buttons are hidden. **Not currently working.** Home Indicator just flashes and fades back to something dimmer. I'm missing something…

### DIFF
--- a/Provenance/Controller/PVControllerViewController.m
+++ b/Provenance/Controller/PVControllerViewController.m
@@ -22,10 +22,15 @@
 @interface PVControllerViewController ()
 
 @property (nonatomic, strong) NSArray *controlLayout;
+@property (nonatomic, assign) BOOL shouldHideHomeIndicator;
 
 @end
 
 @implementation PVControllerViewController
+
+-(BOOL)prefersHomeIndicatorAutoHidden{
+	return self.shouldHideHomeIndicator;
+}
 
 - (id)initWithControlLayout:(NSArray *)controlLayout systemIdentifier:(NSString *)systemIdentifier
 {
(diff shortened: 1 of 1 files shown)


jasarien:
A possible reason why this might not be working is that it may conflict with the deferred screen edge system gestures: https://github.com/jasarien/Provenance/blob/ddfeefd56decd8cd4b82665f9ee471947f7db51f/Provenance/Emulator/PVEmulatorViewController.m#L390-L393

I think one overrides the other. You’d need to disable one in favour of the other in each situation.

sevdestruct:
@jasarien Which one is getting the Home Indicator drawn to it the PVControlerViewController or PVEmulatorViewController that is under it. It seems that the PVConvtrollerViewController has it from what i can tell. I read some documentation that says if it's a child view, than there is an extra bit? is PVConvtrollerViewController a child view of anything? 

…unless PVEmulator also is receiving it, so when i see it fade it's actually doubled up, so it reveals the EmualtorView's Home Indicator?

jasarien:
Yes, the controller view controller is a child view controller of the emulator view controller.

sevdestruct:
Thanks. Still can't seem to get it. Tried testing disabling edge gesture piece you isolated, and also tried adding a bit for child view following solution for this stackoverflow: [hide home indicator for iPhone X](https://stackoverflow.com/questions/47322175/hide-home-indicator-for-iphone-x) that documents it in Objective-C. There's also some decent discussion around it in Swift in a Medium post [iPhone X: Dealing with Home Indicator](https://medium.com/the-traveled-ios-developers-guide/iphone-x-dealing-with-home-indicator-2e8e47f5647f) that explains it in more detail…

sevdestruct:
Wait, think I may have got it!

PR #677, opened 2018-02-05
N64 support (mupen core)

braindx:
I've taken the awesome work of @JoeMatt integrating the mupen core into Provenance and took it the rest of the way toward getting it functional. Primarily, this involved adding support for cores that render directly to the GPU via OpenGL. This new render path renders to double-buffered textures and allows the rendering to take place in parallel on another thread. Additionally, I made some changes to force use of the pure interpreter (disabled dynarec and the cached interpreter), disabled some compiler optimizations in the interpreter that were causing issues, fixed up input (controller overlay, MFi gamepads, etc.), added support for restarting audio (needed for when N64 changes sample rates), fixed up deadlocks when loading/saving states, and fixed up a bunch of places where mupen didn't clean up properly after itself when shutting down.

I think this puts N64 emulation in a pretty decent state. Future work would likely include getting alternate plugins working like gliden64 to improve compatibility as I'm unsure what games work and how well.

PR #685, opened 2018-02-14
Feature/n64+swift

JoeMatt:
This contains all the changes from #677 plus,

- WebDav server for copying ROMs (always on setting added for aTV)
- tvOS TopShelf sizes added for other cores
- Realm updated - Swift version
- Additional metadata added to database
- Game Info screen for iOS and aTV to show additional metadata
- Convert to Swift - all aTV classes. Game Library, most of settings views and model
- iOS 3D touch on springboard shows favourites then recents
- Full spotlight indexing
- Improved logging
- Updated Opening db to 24.0
- Custom volume bar while in game
- Build script no long touches Info.plist in source directory, instead modifies in build directory to prevent git wanting to commit new version
- Improved error handling all around
- UIUserActivity fixed, other devices show shortcut when web server running
- Fix a bunch of stuff that would have crashed iOS 8 or 9
- Added L2/R2 support
- Added m3u multi-disc support for PSX
- Added UIDocumentPickerViewController to allow importing games from iCloud and 3rd party apps that have a document picker extension. This also allows importing from NAS drives in-app when paired with 'FileExplorer Pro' from the app store after setting up NAS connections in that app.

Images of some of the new UIs can be seen here,
https://imgur.com/a/lAjd2

Bugs to fix:

- [x] Import of CD systems
- [x] Test L2/R2 after CD is fixed
- [x] iCade Settings on tvOS
- [x] Importer duplicating CDs
- [x] Handle showing menu and backgrounding app for time in game counter
- [x] Realm for iOS 11.3 / Swift 4.0.2 if available
- [x] Realm Migration : PVRom now uses MD5 as a unique key. Unique keys must be unique, so must implement a migration function to generate MD5s at first load. ROMs that can't should be removed from the database as they are most likely zombie'd entries. A way to present UI while this is happening for large migrations? MD5'ing many Roms could be time consuming at boot.

Please report any bugs in the PR. Once all bugs are fixed I can merge.


ciscosurplus:
Latest build on this branch Import/Export server causes app crash, WebDav accessible but files empty. Apple TV4 11.3, same on Apple TV4K.

jasarien:
Strange... I’m not seeing that crash and I was using the WebDAV just fine. @joe anything you can think of? 

@ciscosurplus we’re you doing anything out of the ordinary?

ciscosurplus:
@jasarien no nothing I wouldn't normally do, I had a build from about a week ago which was working ok (apart from very slow FPS on goldeneye). It could be as I'm running 11.3 on both Apple TVs, and I couldn't build in Xcode 9.3, so built in 9.2 then manual upload (as 9.2 doesn't have 11.3 support file). 

WebDav I can connect but I don't see any folders and can't upload anything. I will delete re-build and try again with some de-bugging on.

jasarien:
Hm, I haven’t tried it on the beta, could be something has changed there.

ciscosurplus:
I will pull my non 4K back to 11.2 and test again.

ciscosurplus:
Ok, I got the same crash on 11.2.5, I have the log file for the crash attached. 
[Prov_Crash_16_Feb.log](https://github.com/jasarien/Provenance/files/1732958/Prov_Crash_16_Feb.log)


I build a Fork using  Muppen-Rendering on 8th Feb this worked fine. 


```
Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x0000000000000010
VM Region Info: 0x10 is not in any region.  Bytes before following region: 4304175088
      REGION TYPE                      START - END             [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      __TEXT                 00000001008c8000-00000001009e0000 [ 1120K] r-x/r-x SM=COW  ...p/Provenance]

Termination Signal: Segmentation fault: 11
Termination Reason: Namespace SIGNAL, Code 0xb
Terminating Process: exc handler [0]
Triggered by Thread:  0

```

ciscosurplus:
Have to say some initial testing on new commit, issue with crash/server (I raised above) is fixed. 

Also I tested Goldeneye (very intensive game) it runs very smoothly on latest build on ATV4 and sound plays fine with minor skip now and then. 

Great work thank you, will do some more testing and report back. More than happy to run specific test/ help in any way I can. 

2 minor things

1. When attempting to Rename a file, as soon you type any letters it crashes out of the app. 
2. it won't build in Xcode 9.3 Beta (needed for TVOS 11.3 device loading) - I only mention is this is due to go public in 2-3 weeks import RealmSwift"Module compiled with Swift 4.0.2 cannot be imported in Swift 4.1:" but I guess the easy fix here is just for me to manually replace the frameworks.

JoeMatt:
@ciscosurplus There's nothing we can do about the beta. Realm is shipped as a binary framework and Swift 4 api is still fragile so they need to prove a new binary built against the newest version of Swift before we can fix that. Swift 5 is supposed to stabilise the ABI and fix that but that's down the road.

I'll look into the rename crash.

ciscosurplus:
@JoeMatt figured thats the case, it was more of a heads up for public release. 

I noticed a few other minor issues on TVOS

1. Click on icade setting and all the settings disappear (displays a blank list)
2. Auto save on several N64 games causes it to lock up (screen locks up sound continues) I will try to debug this but disabling auto save seems to clear the issue. 

Will test multi-disc swapping this evening.

abesada:
Platform test : appleTV simulator and device

The compilation is successful but crash when I execute the app:

Fatal error: Migration is required due to the following errors:

    Property 'PVGame.romSerial' has been added.: file .../Provenance/Provenance/Game Library/Database/RomDatabase.swift, line 139

JoeMatt:
If you were testing a previous build you have and you get a migration error you have to delete the app, or increase the DB number for realm. It should install over a the prior official builds though, I think. If it' snot let me know.

ciscosurplus:
@JoeMatt Ive started to test multi-Cd (resident evil 2) it loads up with the M3U and due file, however Im having a separate issue in that I can't hit start so I can't test swapping CD until I can actual start a game. I posted in the PSX controller section... will keep you posted.

ciscosurplus:
Seperate comment re start button, not working on Apple TV (for PSX Anyway) works fine on IOS, tested on hardware TVOS 11.2.5 and in simulator, its not sending any output in the debug for some reason.

I tested on the Master and it works fine on simulator and real.

ciscosurplus:
The icade controller setting is an easy fix on TVOS, you have Target Membership for Vi Cade Controller View Controller as unticked (for TVOS), if enabled for Target Membership it works fine. 



"2018-02-19 13:16:32.655023+0000 Provenance[26115:1295281] Unknown class PViCadeControllerViewController in Interface Builder file."

ciscosurplus:
Cant for the life of me get the context menu P1 Start button to issue a start in this build, can someone point me to the file that handles that menu and command in Xcode?  I worked around by adding a button combo to issue a start command (L1 + R1 + x) but not ideal.

jasarien:
My only suggestion would be to ensure that iCade mode is disabled. And kill/relaunch the app so you know its disabled from the outside and try again — iCade interferes with MFi controllers on tvOS (though, the P1 Start functionality shouldn’t be affected, since it’s just a call to tell the emulator core to press start, the controller shouldn’t be involved), but I’m grasping at straws now, since this definitely works for me 100% with a Nimbus and a Stratus on iOS and tvOS.

ciscosurplus:
@jasarien confirmed its disabled, Im getting the same thing running it in 11.2 simulator mode as on a real box. Are you running 11.2/11.2.5? I have a Nimbus in the garage somewhere will dig it out as currently on physical box using Mad Catz and give that a try. 

As I mentioned Its only on this branch, Master I don't have the issue (in sim or real) Im hoping it just that particular controller and not a combo of the build and TVOS version, but that wouldn't explain the same behaviour in simulation mode running 11.2. Cant comment on iOS.

jasarien:
I’m running it on iOS 11.2.5 — I can’t recall exactly what version my AppleTV is running, I’ll confirm later. As I said earlier, the actual controller shouldn’t matter. The code is here: https://github.com/jasarien/Provenance/blob/feature/N64%2BSwift/Provenance/Emulator/PVEmulatorViewController.m#L471:L474 — All it’s doing is telling the current controller view controller to press start, and then release it half a second later - it should have no bearing on the physical controller itself.

ciscosurplus:
Ok found my Nimbus charged it and tested on both AT4 and ATV4K and same problem maybe it’s the individuals games or a corruption in my build but have been re building as new builds as they come out. Tested on 11.2.5 and 11.3 tvOS. 

For now have worked around it by creating a button combo in MednefenGameCore.mm 

Maybe ignore if I’m the only person seeing the problem.

ciscosurplus:
Rename crash if it helps is inside;

`PVGameLibraryViewController.swift

    @objc func handleTextFieldDidChange(_ notification: Notification) {
        if let text = searchField.text, !text.isEmpty {  **`Thread 1: EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP, subcode=0x0)`**

         searchLibrary(text)
        }
        else {
            clearSearch()
        }
    }`

tonycoco:
Getting some errors:

```
Thread 1: Fatal error: Migration is required due to the following errors:
- Property 'PVGame.romSerial' has been added.
- Property 'PVGame.lastPlayed' has been added.
- Property 'PVGame.playCount' has been added.
- Property 'PVGame.timeSpentInGame' has been added.
```

JoeMatt:
It's in RomData.swift. I added a unique key to the entries. I'm writing the migration right now.

If uyou're going from Beta to Beta though you either have to manually bump the number for now at the top of that file, or start from a clean install.

tonycoco:
Ahh, thanks. I'll manually bump. Also note that it doesn't seem that the `Cover Art` folder works to import in matches to a game like it used to for me on tvOS (ATV4K). It disappears from the folder but never updates the matching ROM.

JoeMatt:
The importer is a work in progress. It's noted in the top comment what's still TBD.

tonycoco:
From: `PVTVSettingsViewController`... `Nil is not compatible with expected argument type 'PVGameImporterCompletionHandler' (aka '(Bool) -> ()')`

sevdestruct:
image uploads working again - confirmed (note: now works from roms folder as well as cover art folder) - @joe suggesting renaming 'roms' folder to 'Imports' (has my vote) - if so, should we lose the cover art [import] folder?

dallasgroot:
Is this why I am getting build errors with the latest version???

Issue #700, opened 2018-02-27
Crash when import file from Onedrive.


genkino:
* This is not a common situation. :)
* from : PVGameImporter.m > lookupInfoForGame
    * fileName is nil (See attached image)
* Logs
> 2018-02-27 12:51:38.977609+0900 Provenance[766:321842] [DYMTLInitPlatform] platform initialization successful
> 2018-02-27 12:51:39.733659+0900 Provenance[766:321501] [MC] Lazy loading NSBundle MobileCoreServices.framework
> 2018-02-27 12:51:39.734483+0900 Provenance[766:321501] [MC] Loaded MobileCoreServices.framework
> 2018-02-27 12:51:40.307202+0900 Provenance[766:321857] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[NSPathStore2 substringToIndex:]: Index 18446744073709551615 out of bounds; string length 21'
> *** First throw call stack:
(0x181af3164 0x180d3c528 0x181af30ac 0x1823fb6c4 0x104485a94 0x104485430 0x104475820 0x106d312cc 0x106d3128c 0x106d3ff80 0x106d347ec 0x106d40f6c 0x106d48020 0x18171af1c 0x18171ab6c)
libc++abi.dylib: terminating with uncaught exception of type NSException
(lldb) 

<img width="573" alt="2018-02-27 12 53 10" src="https://user-images.githubusercontent.com/6399050/36711835-e66c4724-1bc8-11e8-8042-94ef69eb1746.png">

Viswesh934:
assign me this issue

JoeMatt:
Closing: this is a raw crash log from 2018 with no reproduction steps or context. If you still experience crashes importing from OneDrive, please reopen with steps to reproduce, device model, and iOS version.

PR #701, opened 2018-02-27
fixing crash when import file from onedrive.

genkino:
refs #700


JoeMatt:
This pull request is against Master but Master doesn't have support for import from cloud sources yet. I'm assuming you're testing against the Swift branch I'm working on. The PR should be against that branch.

JoeMatt:
Actually I'm just going to close this PR. It's not going to work in the new Swift branch at all since that method has been rewritten to deal with nulls, and if an import got this far with a null path that's bad anyway. 

If this is still an issue in the N64+Swift branch it can be fixed higher up in the import process.

PR #708, opened 2018-03-03
Minor bugs

nadavcoh:
Some minor bug fixes


JoeMatt:
I'm going to close this because PVGameImporter.m is completely re-written in Swift now in my N64+Swift branch. I went ahead and added then ',' to the char set in that branch in commit e1e66cf

PR #738, opened 2018-03-09
Develop merge-down with UIRedesign

sevdestruct:
grabbing Joe's latest commits


sevdestruct:
So the top of this PR says this, which is what i was trying to do... 
screen shot 2018-03-08 at 10 45 36 pm

Yes the bottom says…
screen shot 2018-03-08 at 10 46 45 pm
Which is not what I was doing.. Now I am wary to merge this..

sevdestruct:
Nevermind, seems joe already committed changes from me anyhow.

PR #768, opened 2018-03-27
Update PVGLViewController.m to decouple Image Smoothing from Fragment shader.

mrjschulte:
Restores the previous "Image Smoothing" behavior from being always enabled and softening the results when using the CRT Shader. The feature can now be applied separate from the CRT Shader allowing for explicit pixel accurate rendering without filtering when using the Fragment shader render path in Provenance.


JoeMatt:
I added this in 4bde9ba along with caching the render settings to avoid extra function lookups and calls in the render loop. KVO should work if the settings are changed during running, which isn't support yet but will be down the road.

PR #772, opened 2018-03-28
Add favorites and recently added to TopShelf

entourloop:
Complete the TopShelf results using favorites and recently imported games

### DIFF
--- a/TopShelf/PVGame+TopShelf.swift
+++ b/TopShelf/PVGame+TopShelf.swift
@@ -1,0 +1,47 @@
+//  PVGame+TopShelf.m
+//  Provenance
+//
+//  Created by entourloop on 2018-03-29.
+//  Copyright © 2015 James Addyman. All rights reserved.
+//
+
+import TVServices
+
+// Top shelf extensions
+extension PVGame {
+    public func contentItem(with containerIdentifier: TVContentIdentifier) -> TVContentItem? {
+
+        guard let identifier = TVContentIdentifier(identifier: self.md5Hash, container: containerIdentifier),
+        let item = TVContentItem(contentIdentifier: identifier) else {
+            return nil
+        }
+
+        item.title = self.title
+        item.imageURL = URL(string: self.customArtworkURL.isEmpty ? self.originalArtworkURL : self.customArtworkURL)
+        item.imageShape = imageType
+        item.displayURL = self.displayURL
+        item.lastAccessedDate = self.lastPlayed
+        return item
+    }
+
+    var displayURL: URL {
+        var components = URLComponents()
+        components.scheme = PVAppURLKey
+        components.path = PVGameControllerKey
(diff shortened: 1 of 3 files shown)


JoeMatt:
Could you provide  a screen shot or pic, it's much quicker than checking out a PR and building etc

JoeMatt:
Also looks like you're  using a single method with an early return if a category is empty. Would be better code to make methods to return the content groups and use a smaller function to group and return for maintainablity

TopShelf/PVGame+TopShelf.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    var displayURL: URL {
+        var components = URLComponents()
+        components.scheme = PVAppURLKey
+        components.path = PVGameControllerKey
+        components.queryItems = [URLQueryItem(name: PVGameMD5Key, value: self.md5Hash)]
+        return (components.url)!
+    }
+
+    var imageType: TVContentItemImageShape {
```
JoeMatt: Since this code already exists in PVRecent, it makes more sense to make a single extension of PVSystem with 'var imageType' and refactor the PVRecent extension to call `game.system.imageType`.

If it's like this then when we add a new enum value that's just one more place we have to make sure the switch statement is exhaustive and it's duplicate code

TopShelf/ServiceProvider.swift
```diff
@@ -36,26 +36,68 @@ public class ServiceProvider: NSObject, TVTopShelfProvider {
...
             }
-            recentItems.title = "Recently Played"
-
-            let database = RomDatabase.sharedInstance
-
-            let recentGames = database.all(PVRecentGame.self, sortedByKeyPath: #keyPath(PVRecentGame.lastPlayedDate), ascending: false)
-
+            favoriteItems.title = "Favorites"
+            let favoriteGames = database.all(PVGame.self, where: "isFavorite", value: true).sorted(byKeyPath: #keyPath(PVGame.title), ascending: false)
+            var favoriteNames = [String]()
```
JoeMatt: `database.all(PVGame.self, where: "isFavorite", value: true).sorted(byKeyPath: #keyPath(PVGame.title), ascending: false).map {...' could completely get rid of the for loop and mutable array

entourloop:
This early implementation is pretty naive, I'll do some refactoring a bit later. I attached a screenshot below to illustrate the feature.
screen shot 2018-03-29 at 07 25 05

sevdestruct:
@entourloop, how’s your refactor coming?

JoeMatt:
It's such a simple change I could just redo it myself while i wait for this dumb thing to compile

entourloop:
The requested refactoring was done, but titles unicity (e.g. not having a game appearing in "recently added" when it's been shown in "favorites" first) has been removed, for better code readability.

entourloop:
I also had some issues at first with the Top Shelf extension not updating itself, or not running. IIRC, I had to test the “TopShelf” target directly so the changes apply. I also updated my Apple TV, just to be sure

PR #782, opened 2018-03-31
New UI for Save States

jasarien:
A new UI for save states.

- Auto save states no longer overwrite the previous state, so you can load back through your history of auto saves, as many auto saves as you like.
- Manual states are no longer limited to 5 slots. As many manual save states as you like.
- Save states now show a screenshot of the state
- Old save states are converted to the new system, so you shouldn't lose your old ones when this gets merged and you update
- Tap and hold (or press and hold A on a controller on tvOS) to delete save states.


JoeMatt:
Trivis is gonna fail I think on Swift 4.1. We should probably stick to 4.0 for now and ignore the warnings.

PR #789, opened 2018-04-02
Fixes crash that can happen when launching a game.

MichaelMKenny:
I told @Jasarien on discord about a crash when launching a game, and proposed a fix. He then committed the fix (379023f4). Unfortunately, I found it still crashes in some circumstances and have come up with a better fix. Also, I applied the fix to not just Recently Saved, but also to Recently Played.

Btw, this is my first Pull Request ever.

----

The Recently Saved section only displays the last 6 items, but the user probably has way more states saved than that. So when the collectionView sections update, sometimes the Realm update notification wants to update items outside the first 6, but those items don't exist in the collectionView. This causes a crash.

To fix this, we filter the deletions, insertions, and modifications to only those within the first 6 items.

### DIFF
--- a/Provenance/Game Library/UI/PVGameLibraryViewController.swift
+++ b/Provenance/Game Library/UI/PVGameLibraryViewController.swift
@@ -459,7 +459,7 @@ class PVGameLibraryViewController: UIViewController, UITextFieldDelegate, UINavi
 				}
 
 				// Query results have changed, so apply them to the UICollectionView
-				self.handleUpdate(forSection: section, deletions: deletions, insertions: insertions, modifications: Array(modifications.prefix(self.maxForSpecialSection)), needsInsert: needsInsert, needsDelete: needsDelete)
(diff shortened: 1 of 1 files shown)


jasarien:
Awesome, thanks @MichaelMKenny we'll get it tested

jasarien:
This still crashes for me when you delete a save state. I believe the issue is that we have more than 6 save states, and when you delete one, it tells the collectionView there is 1 deletion, but the data source is still going to return 6 states for the section, because another takes it's place. So the exception is thrown that 1 was deleted, it was expecting 5, but got 6... We probably need to detect the case when a deletion is going to get backfilled, and transform it into a modification rather than a deletion

jasarien:
So I merged your changes, and then I added this commit 5767249ef273355f24e858a5b7c5a113938963ac on top to help with the extra crashes I saw. I think its a lot better now.

PR #836, opened 2018-04-15
Update crt_fragment.glsl

mrjschulte:
Fix SCANLINE_MIN_BRIGHTNESS to remove "pinkness" by setting the vec3 to 0.25 as discussed w BrainDX.


mrjschulte:
Builds fine here,

Issue #803, opened 2018-04-09
Start/Select/Shoulders/Triggers Vibrator Turbo glitch


sevdestruct:
Tap and hold on any of these (they are of same class) and swipe off, staying held, and the haptics will go into turbo mode.

sevdestruct:
Fixed.

PR #837, opened 2018-04-16
Issue #803: tweak logic to check touch area intersection to avoid repeated …

yoshisuga:
…buttonPressed delegate method calls on touchedMoved

This affects buttons that are not part of the PVButtonGroupOverlayView, including start, select and shoulder buttons.

This addresses #803


jasarien:
@gh0sti  No, that is a different issue.

sevdestruct:
Confirming this fixes #803 (turbo vibrator glitch), also using 'closes #803' in your PR line should auto close the ticket on merge (as it does when we push)

PR #873, opened 2018-04-29
Update MednafenGameCore.mm

mrjschulte:
Resets PCE/PCE_FAST Scanline counts back to defaults in MDFNI. Fixes the broken pixel aspect in games and removes edge of frame items one would never see on a CRT.


JoeMatt:
Looks good, I'll merge it and take your word for it and it works

PR #890, opened 2018-05-12
Save state tweaks 

yoshisuga:
Dismiss save state controller immediately after saving state; show notification banner when state loaded/saved/overwritten

I added a simple notification banner class from:
https://github.com/bryx-inc/BRYXBanner

I was thinking of implementing quick-load and quick-save states, and I thought having this simple notification banner would be useful and provide some much needed feedback.

I also thought it was more convenient to get right back into the game when saving/overwriting states.

### DIFF
--- a/Provenance.xcodeproj/project.pbxproj
+++ b/Provenance.xcodeproj/project.pbxproj
@@ -170,6 +170,10 @@
 		42BC83A917E6775E00E9A607 /* PVGameLibrarySectionHeaderView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42BC83A817E6775E00E9A607 /* PVGameLibrarySectionHeaderView.swift */; };
 		42E63F0617DE78AB005802B0 /* UIImage+Scaling.m in Sources */ = {isa = PBXBuildFile; fileRef = 42E63F0517DE78AB005802B0 /* UIImage+Scaling.m */; };
 		7556CF7A19DC15C1007F8F97 /* Default.xib in Resources */ = {isa = PBXBuildFile; fileRef = 7556CF7919DC15C1007F8F97 /* Default.xib */; };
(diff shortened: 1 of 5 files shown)


JoeMatt:
Not sure I'm a fan of that banner class, seems really hacky. Tested on tvOS? 

Also this commit contains team changes to project file.


Having a hard time following what this change is exactly. Can you put step by step differences from current behavior.

sevdestruct:
@yoshisuga this might be a bit much for save states for 2.0 i am thinking of doing more of an autosave icon animation, discreetly, in concert with the mini HUD noted in my UI 2.0 Trello cards. This type of banner seems more useful for something else, like active web server or something.

yoshisuga:
I'll need to test on tvOS, not actually sure if it'll work there. I thought we can use this banner class since it's just 1 class, no dependencies, and it seems to do what it says it does - use autolayout constraints to animate a view on top of a window or view.

So the main point of this PR is actually really simple: to go back into the game right after saving your state. 

Current behavior:
- In Save States screen, tap "+" or tap on existing save state and overwrite
- The save state gets added to the save states collection view, or replaced in the collection view
- Dismiss the save states screen to go back into the gam

New behavior:
- In Save States screen, tap "+" or tap on existing save state and overwrite
- Save states screen is dismissed right away and you get back into the game

The banner stuff is just a way to provide additional visual feedback to indicate what happened - a replacement of the collection view insert/delete animation that happens. It's really just a small quality-of-life improvement that i think would be welcomed.

I'm fine with ditching the banner class and just dismissing the save states controller after creating/overwriting a new save state, esp since @sevdestruct has some plans for the future UI. Does that sound ok?

Oh, and yeah I'll revert that team change in the project file :) Thanks for the review :)

sevdestruct:
the return to game after saving state is fine on it's own, decouple it from this PR, and submit it separately, also, remove the files you shouldnt change from the commit (xcode project files etc) that were modified with your Bundle ID, and so on.

sevdestruct:
@JoeMatt is currently working on save state, etc… so I'll let him review this and incorporate if it hasn't already been done in his uncommitted branch.

yoshisuga:
See #892 - closing this one.

PR #892, opened 2018-05-14
dismiss save state controller immediately after saving/overwriting state

yoshisuga:


JoeMatt:
Thanks, I'll take a look at this after I get my save state work completed.

JoeMatt:
I'm going to reject this PR but I'll give why,

1. Not sure what benefit this really adds
2. View Controllers usually shouldn't self dismiss but instead let their presenting parent decide when because 
3. This method throws, if the parent decides to present an error message and this VC self dismisses without waiting, the alert would be presented at state that would dismiss the alert, not the save state VC, which 
4. If this method does throw, the dismiss won't be called either

As you can see, a lot of testing and conditions for something that just saves a back button tap.

PR #909, opened 2018-05-22
Add “Files” support to info.plist

konistehrad:
Allows navigation to file system structure via "Files" interface. Sorta convenient, as it allows copying saves and ROM's to and from iCloud Drive.


konistehrad:
Example images of using Files interface:

img_1897
img_1898
img_1899

sevdestruct:
this is already supported and fully functional in 1.5, not sure what this is for…

konistehrad:
Ah my apologies! Thanks for the quick response.

PR #969, opened 2018-07-18
Fixing PSX Analog Controller for MFI.

Dybuk:
- Switching gamepad to dualanalog for input.
- Sending correct input analog input for Mednafen in didMovePSXJoystickDirection.

------

I switched to dualanalog rather than dualshock due to the 'always on' analog (removing the need for the separate button on the UI), the simpler analog maths and the lack of L3 and R3 (thanks apple!).

These are my concerns about this change:

- Its not all too well tested. I tested on "ace combat 3" and metal gear solid.
- I'm not sure if switching from gamepad to dualanalog for all controllers is the best approach. Maybe its better to only switch to dualanalog when using an MFI controller. Or maybe go back to adding that analog button.
- The code might not be as optimised as possible and it gets called a lot. Maybe there is scope to improve this.
- There is a mention in the comment there about square vs circle axis ranges. I have not added any additional maths to add a fudge factor for this.

Maybe these are good talking points.

Many thanks for your work on this project. It is appreciated.


jasarien:
This looks like a good start, thanks for submitting this PR!

I’ll try to test this tonight, and let you know if anything comes up.

corlettb:
I'm assuming the check failures are because of issues in the develop branch.

jasarien:
Ignore the travis checks... It times out because the build takes too long....

Dybuk:
I still haven't done anything with L3 and R3 yet.

JoeMatt:
Yeah I keep trying to fix travis but it hates me, and you have to make a change, upload and see if it worked. you can run travis in a docker but that's way too much work for me so, it is what it is.

re the PSX changes. Looks good at first glance, I'll take a closer look in the coming days when I get some time.

Dybuk:
I think I spotted an issue with loading a save state from before these changes were applied when just using the overlay. I think the game picks up diagonal input that you can’t correct without adding an mfi controller. I’ll try to reproduce it and fix.

I’m still not sure what to do on the L3 R3 front.

If I think of modifications just for me I’d probably move the mfi overlay buttons. I’m using a grip controller on an iPad.

Move the start and select out of the bottom corners into the middle left and right so they can be accessed easier. Then underneither them add l3 r3. 

I might even put L3 on the right and R3 on the left. I know as a modern gamer that L3 is often expected to be clicked as you are pushing in a direction, like to sprint. I’m not sure how often this is the case on the PS1 though.

I bet not everyone would agree with my layout.

I know you guys are working on key combos... but if you use an existing key for the modifier then you might jump in game... shoot off one of your precious bullets etc.

You could try to use the menu button as a modifier and then try to pair that with the shoulder buttons and DPad for different options.

Curse you Apple.

Dybuk:
I've added l3 and r3 overlay buttons only visible when:

- a controller attached
- playing a psx game
- the option "start/select always on-screen" is enabled.

These should appear above the start/select buttons.

I do realise that I may have shot my own pull request in the foot by messing with the overlay code. There be dragons there.

I was also quite cheeky and added a gripControl constant which can be updated to move the controls into a better position if playing with a grip controller.

I'll give it more of a test over the coming days.

JoeMatt:
Merged and added a modifier combo for mfi, 4 shoulders and A+X so it would work on Apple TV

Issue #679, opened 2018-02-07
Modifier Button Combos to extend MFi controllers/gamepads (MFi+)


sevdestruct:
In response to the discussion in #644 (upgrading this to a larger topic for all existing cores and new additions, not just PSX) and until we have button customization, #56 implemented…

Been thinking of a solve for buttons outside of the MFi controller default to get Start, Select (many systems) L3, R3 (PSX) and potentially C buttons (N64) accesible on MFi controls like SteelSeries Nimbus or Gamevice which is what I am using (though it seems the SteelSeries Stratus XL has a couple extra buttons and also now supports iOS (potentially tvOS as well) if you update the firmware - if that becomes the recommended gamepad of choice, and if those extra buttons are reachable…)

What if we did something like this (based on Nimbus layout button names), think of the L/R button combos as modifiers:

Select = L1+L2 + DpadDown
Start = R1+R2 + A

…or to get @jasarien mention of L3/R3 in:

Select = L1+L2 + DpadRight
Start = R1+R2 + X
L3 =  L1+L2 + DpadDown
R3 = R1+R2 + A

so the controllers are at least in a similar relation to their actual positions or relation to each other.

If that combo ends up being something to easy to accidentally press, then how about requiring all 4 L/R buttons instead of just 2 for the above suggestions (although I was thinking maybe to save those combos for N64 C buttons unless we can utilize the L/R for those for those…i forget how many buttons N64 has… only one trigger (Z) right (preference of L or R, default R…)? so that leaves us with 3 extra triggers for some of the main used C buttons and/or some form of L1L2R1R2 + ABXY) 

Trying to arrive at something that could be fairly universal across the different systems. Thoughts?

Perhaps, we can show something subtle on screen when the combo sets are pressed to inform users about the temporary virtual button mode they are in until they release the requisite  combo sets (the L/R combos as modifiers).

JoeMatt:
Certain games on certain systems are going to require different sets of combo's, if that's the way to go.

Honestly this could all be solved by just adding a custom controller mapping setting per core that supports combos. It's not a difficult feature to write, but it would be time consuming and tedious since you'd have to go into all of the current cores and re-write how they handle events. There's not a consistent protocol that they all follow. Plus adding UI, even if somewhat basic.

I mean, this will probably come but no one is exactly raising their hands to write it either. We don't even have CD switching support wired up yet, which is something I recently promised to do.

[...]

PR #978, opened 2018-08-05
Adding dpad emulation to left stick when psx not in analog mode

Dybuk:
This change will add dpad emulation to the left analog stick for PS1 similar the other cores. For PSX we only do this when the analog mode is disabled.

Give me a couple of days to test this.

I might also want to consider adding an option to disable the key combos. Games like Wing Command 3 use a lot of their own combos and these blocked. 

https://www.wcnews.com/controls.shtml#wc3ps

### DIFF
--- a/Mednafen/MednafenGameCore.mm
+++ b/Mednafen/MednafenGameCore.mm
@@ -820,10 +820,12 @@ static void emulation_run(BOOL skipFrame) {
         }
         
         if (controller) {
+            uint8 *d8 = (uint8 *)inputBuffer[playerIndex];
+            bool analogMode = (d8[2] & 0x02);
             for (unsigned i=0; i<maxValue; i++) {
-
+                
 				if (self.systemType != MednaSystemPSX || i < PVPSXButtonLeftAnalogUp) {
-                    uint32_t value = (uint32_t)[self controllerValueForButtonID:i forPlayer:playerIndex];
+                    uint32_t value = (uint32_t)[self controllerValueForButtonID:i forPlayer:playerIndex:analogMode];
                     
                     // TODO Can we do this better?
                     // we don't want to read l3/r3 from the controller
@@ -837,7 +839,7 @@ static void emulation_run(BOOL skipFrame) {
                     } else {
                         inputBuffer[playerIndex][0] &= ~(1 << map[i]);
                     }
-                } else {
+                } else if (analogMode) {
                     float analogValue = [self PSXAnalogControllerValueForButtonID:i forController:controller];
                     [self didMovePSXJoystickDirection:(PVPSXButton)i
                                             withValue:analogValue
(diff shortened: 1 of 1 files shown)


sevdestruct:
looks like you missed implementing this for standardGamepad as well. Should be right below the extended set

sevdestruct:
Thanks for the heads up on WC3, can you add that note and link to #679 to keep in mind the later iterations of MFi+ Mode

Dybuk:
By the standard gamepad do you mean the non-extended version of the mfi controller? I thought that didn't have analog controls. Maybe I got that wrong. I'll take a look.

jasarien:
@Dybuk you’re right, only extended controllers have analog sticks. Standard doesn’t have them. 

It also seems like Apple are putting less emphasis on the standard profile — it’s not even mentioned in their programming guide for the game controller framework  and I think nearly all MFi game controllers still being sold today are extended profile.

Dybuk:
I guess I could add it to the standard gamepad incase in the future we have an analog stick in the overlay controller.

sevdestruct:
Right. Not sure why I mentioned. Hah, my bad. standard has no sticks. Ignore what I said.

sevdestruct:
Testrun complete, works for me.

Mednafen/MednafenGameCore.mm
```diff
@@ -1363,7 +1365,7 @@ - (void)didReleaseWSButton:(PVWSButton)button forPlayer:(NSInteger)player;
     inputBuffer[player][0] &= ~(1 << WSMap[button]);
 }
 
-- (NSInteger)controllerValueForButtonID:(unsigned)buttonID forPlayer:(NSInteger)player {
+- (NSInteger)controllerValueForButtonID:(unsigned)buttonID forPlayer:(NSInteger)player:(bool)analogMode {
```
JoeMatt: This is missing a parameter label, I'm not even sure how t his

Mednafen/MednafenGameCore.mm
```diff
@@ -2038,13 +2040,31 @@ - (NSInteger)PSXcontrollerValueForButtonID:(unsigned)buttonID forController:(GCC
         bool modifiersPressed = modifier1Pressed && modifier2Pressed;
         switch (buttonID) {
             case PVPSXButtonUp:
-                return [[dpad up] isPressed];
+                if (analogMode) {
+                    return [[dpad up] isPressed];
+                }
+                else {
+                    return [[dpad up] isPressed]?:[[[pad leftThumbstick] up] isPressed];
```
JoeMatt: Why the use of the ternary here? A short circuited '||' would do the same thing and more readable.
Really can be 1 line return `[[dpad up] isPressed] || (!analogMode && [[[pad leftThumbstick] up] isPressed]) for all these cases same applies

PR #1040, opened 2018-12-20
Correct Core version reported for Genesis Core.

mrjschulte:
When the Core was updated to 1.7.4 this marker was left at the previous 1.7.1. This minor PR fixes this.


JoeMatt:
I think it's not really a big deal, but technically I'm pretty sure we're not on full 1.7.4. The genplus core is a copy / paste, not a submodule, the last merge was kind of a hybrid as far as I can remember. 

I took a stab and the libretro head, it's close, but there's some refactoring that was breaking and I gave up after a few hours since update genesis isn't a huge deal, though I think that PCM thing that was requested requires it.

PR #1042, opened 2018-12-20
Update Screenshot function to write PNG instead of JPEG

mrjschulte:
Write out PNG's instead of JPEG's for the Screenshot function. Gets us all of the RGBA channels as a 8888 pixel type for much better quality images at a much smaller size. 
Ex. A Genesis Sonic 2 title screen Screenshot at 2048x1536 as PNG is 243KB whilst the original JPEG at 50% lossy compression is 344KB.

### DIFF
--- a/Provenance/Emulator/PVEmulatorViewController.swift
+++ b/Provenance/Emulator/PVEmulatorViewController.swift
@@ -874,14 +874,14 @@ final class PVEmulatorViewController: PVEmulatorViewControllerRootClass, PVAudio
 				UIImageWriteToSavedPhotosAlbum(screenshot, nil, nil, nil)
 			})
 
-			if let jpegData = screenshot.jpegData(compressionQuality: 0.5) {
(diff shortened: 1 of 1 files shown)


mrjschulte:
I'll also add the comment that as the image buffer complexity increases, due to say enabling a CRT shader that makes every pixel "unique" the file size of the PNG's will go up significantly because the lossless compression in PNG doesn't scale as well as the lossy compression in the JPEG files.

If this is a problem do we want to add a PNG/JPEG toggle in the Ui to handle this?

JoeMatt:
These were PNG's, I switched to JPEG's because the PNGs were in many cases actually bigger than the save files themselves due to the lack of any compression options. It might make sense to scale them from the full context size at the very least. Don't need a full retina screenshot when we don't even show the thumbnails full size. Maybe keep screenshots as PNG but I don't really see the need for hi-res lossless thumbnails.

mrjschulte:
Great, then we should be good, right?

PR #1043, opened 2018-12-20
Steam Controller support

zydeco:
This adds support for Steam Controllers in BLE mode.
It works on iOS and tvOS, but on tvOS, the Steam Controller will no longer function as a remote after it has been switched into controller mode, so it can't be used to pause the game or control the UI.


JoeMatt:
I just changed Carthage to use sub-modules. Can you redo this PR with submodule support. You can `git rm -r Carthage/Checkouts/SteamController` then `bundle exec fastlane carthage_bootstrap` and it should recofigure as a submodule. The .framework's will need to be added to the xcfilelists too. You can download the Carting tool, https://github.com/artemnovichkov/Carting , make sure the .frameworks from tvOS (if supported) and iOS are in the link (not embed) frameworks stage, and run `carting update` and it should auto update the file lists.

zydeco:
I had some problem with the submodules, but it seems fixed by cdf38942, so I was able to rebase and add it as a submodule.
I've done some light testing with two Steam controllers and TGBDual on iOS (12.1) and tvOS (12.1.1), and it works as expected (except for the controller mode switching problem on tvOS mentioned above).

sevdestruct:
I have a steam controller. I can see about giving this a test run on my end, soon @zydeco / @JoeMatt

PR #1052, opened 2018-12-31
Switch Mednafen over to pcefast core

mrjschulte:
Enable all Pce rendered scanlines for full raster support and default back to allow for run+select “soft reset” like on the real console. Leave CD-Rom Read speed tweak to 4x as is.

### DIFF
--- a/Cores/Mednafen/MednafenGameCore.mm
+++ b/Cores/Mednafen/MednafenGameCore.mm
@@ -45,7 +45,7 @@
 #import <PVSupport/PVSupport-Swift.h>
 
 
-#define USE_PCE_FAST 0
+#define USE_PCE_FAST 1
 #define USE_SNES_FAUST 1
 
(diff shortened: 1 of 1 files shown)


JoeMatt:
I think there was a reason I reverted from PCE fast to PCE, which is also what OpenEMU uses (not that they're the authority since they've had bugs before I've fixed). Waiting for feedback from the testers we had on PCE a while back (I have little experience since I don't speak Japanese which is like 95% of the games)

mrjschulte:
Yeah, that was me too m that revert originally. The update to Mednafen 1.21.3 back in November helped stabilize the pce_fast module somewhat and it also looks to be the more active module in development. Also, pce_fast is the only module that will run for lower powered A7 iPad Air 1 HW w the CRT shader because of that silly framebuffer size that the regular pce module passes to GL.

JoeMatt:
@mrjschulte ok cool didn't remember that was you. Well, guess this is safe to merge then.

PR #1057, opened 2019-01-07
Genesis : Add support for Double Buffered rendering

mrjschulte:
Adds in Double Buffer support to the Genesis Core for a smoother experience.


JoeMatt:
I'm not seeing where this core is actually double buffered. Snes9x does use 2 buffers with a swap, but `PVGenesisEmulatorCore` only has a single buffer allocation. Where's the swap?

JoeMatt:
closed in merge 089bce4e4cc25e134ee3843bdceb92976dd5262e

PR #1059, opened 2019-01-13
Change order of gb cores

error404-na:


error404-na:
Order is now Gamebatte, Mednafen, TGBDual

PR #1078, opened 2019-02-07
Adding 'bin' to PSX's PVSupportedExtensions

mymikemiller:
Allows Provenance to import zips with bin files

### DIFF
--- a/PVLibrary/PVLibrary/Resources/systems.plist
+++ b/PVLibrary/PVLibrary/Resources/systems.plist
@@ -201,6 +201,7 @@
 		<array>
 			<string>cue</string>
 			<string>ccd</string>
+			<string>bin</string>
 		</array>
 		<key>PVControlLayout</key>
 		<array>


JoeMatt:
`bin` is a reserved extension and no core has it assigned to it. Zips should contain bin/cue's which is the proper way to import. A `.bin` on it's own isn't a valid file for any system other than Genesis which is noted in the Wiki

mymikemiller:
Thank, @JoeMatt! Sorry to bother you with this PR - I meant to delete it last night once I figured out what my actual problem was. I made the PR to show sev the change I made to fix an issue I was having where PSX roms failed to import. He helped me figure out that my rom zip was missing .cue files. 

Oddly, the game worked fine with only the .bins and .m3u (which pointed to nonexistent .cue files) once I manually moved them into com.provenance.psx. Digging around GameImporter, I found that the .bin files weren't being moved. Sorry for not deleting this PR when I realized this wasn't the proper fix. But, hey, now I have a pretty good understanding of how GameImporter works :)

BTW, I'm super impressed with the codebase. You're damn good at what you do!

PR #1083, opened 2019-03-06
Show idle timer setting in tvOS

zydeco:
Not sure why this is not shown in tvOS, the idle timer still does kick in when there's no input, so this adds the option to the settings, like on iOS. It also restores the idle timer to its previous value when stopping the web server, instead of just reenabling it.

I found this when playing for a long time with a Steam Controller, since its input doesn't go through the normal channels, it's not considered input by the OS, I will see if I can solve this on the SteamController framework too, but I think it's only possible with private API.

Would it be better to always disable the idle timer while a game is running? Apple's documentation suggests disabling the idle timer on games that use the accelerometer as input, for example.


JoeMatt:
Im' going to just make it always on for tvOS, no sense really having a toggle for Apple TV since it's not battery powered.

We also don't have notifications for when the setting is change and it's only read at startup which is an issue. I'll make a new ticket to convert settings to RxSwift so they can be observed by anyone.

PR #1176, opened 2019-09-02
tweak precision for blit shader

mrjschulte:
Minor precision tweaks to blit shader for squeezing out extra bit of GPU performance on the lower end of the supported iOS hardware stack such as the iPad Air 1 by setting to lowp instead of mediump/highp.

### DIFF
--- a/Provenance/Utilities/shaders/blit/blit_fragment.glsl
+++ b/Provenance/Utilities/shaders/blit/blit_fragment.glsl
@@ -1,5 +1,5 @@
 #ifdef GL_ES
-precision mediump float;
+precision lowp float;
 #endif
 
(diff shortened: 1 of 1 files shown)


JoeMatt:
Is there any tradeoff for higher devices?

mrjschulte:
Not in my, albeit limited, testing on the iPad Pro 11"

braindx:
Looks good to me

JoeMatt:
A lot of mobile docs for similar mobile GPUs recommend the `mediump` format due to reducing internal stages of precision conversion.

Need to research the Apple chips if this is mentioned anywhere.

Would it be possible to test with all mediump and see if results are similiar?

PR #1177, opened 2019-09-02
prune unneeded gl calls

mrjschulte:
This PR removes the glClear and glDisable calls to help shave some additional ms' off of the GPU render stack on lower end iOS hardware such as the iPad Air 1.

### DIFF
--- a/Provenance/Emulator/PVGLViewController.m
+++ b/Provenance/Emulator/PVGLViewController.m
@@ -491,9 +491,6 @@ struct RenderSettings {
     {
         MAKESTRONG(self);
 
-        glClearColor(1.0, 1.0, 1.0, 1.0);
-        glClear(GL_COLOR_BUFFER_BIT);
-
         GLuint frontBufferTex;
         if ([self.emulatorCore rendersToOpenGL])
         {
@@ -528,10 +525,7 @@ struct RenderSettings {
             glUseProgram(strongself->blitShaderProgram);
             glUniform1i(strongself->blitUniform_EmulatedImage, 0);
         }
-        
-        glDisable(GL_DEPTH_TEST);
-        glDisable(GL_CULL_FACE);
-        
+                
         [self updateVBOWithScreenRect:screenRect andVideoBufferSize:videoBufferSize];
         
         glBindBuffer(GL_ARRAY_BUFFER, strongself->vertexVBO);


mrjschulte:
Together with #1176 these two teaks result in the following performance metrics testing on the Genesis Core with Dragons Fury and using Multi-threaded GL:

**Before:**
Utilization
Tiler - 37%
Renderer - 34% 
Device - 37%
**Frame Time**
CPU - 0ms
GPU  - **6.1ms**

**After** 
Utilization
Tiler - 20%
Renderer - 16% 
Device - 20%
**Frame Time**
CPU - 0ms
GPU  - **3.3ms**

JoeMatt:
Has this been tested to not have GFX glitches on all cores?

Specifically the off thread GL based cores like N64, Reicast.

braindx:
It would probably be better to leave the glClear calls in development builds as it can aid in debugging new cores and video modes/resolutions, but I would think it’s probably fine to skip those generally.

However, I would think it’s possible that skipping the glDisable calls causes issues with certain cores and titles that use OpenGL natively, e.g. certain N64 games, as they may have set up OpenGL state that prevents the subsequent render call from drawing correctly or at all.

JoeMatt:
Sounds like there may be 2 modifications to consider then,

1. #ifdef DEBUG wrap the glClear calls
2. Wrap the glDisable calls in either an `if core.rendersToOpenGL`, or two versions of the function with and without and use a function pointer at Core startup to set the function pointer. This might be an over-optimization to avoid 1 if statement, though would perhaps allow future develop for native and non-native gl render path ways.

Provenance/Emulator/PVGLViewController.m
```diff
@@ -491,9 +491,6 @@ - (void)glkView:(GLKView *)view drawInRect:(CGRect)rect
     {
         MAKESTRONG(self);
 
-        glClearColor(1.0, 1.0, 1.0, 1.0);
```
JoeMatt: As @braindx suggested this might help with debugging,
request wrapping instead in `#ifdef DEBUG`

Provenance/Emulator/PVGLViewController.m
```diff
@@ -528,10 +525,7 @@ - (void)glkView:(GLKView *)view drawInRect:(CGRect)rect
             glUseProgram(strongself->blitShaderProgram);
             glUniform1i(strongself->blitUniform_EmulatedImage, 0);
         }
-        
-        glDisable(GL_DEPTH_TEST);
```
JoeMatt: Instead move to end of above if block since it might be required by native rendering open gl calls.

Most devices running these types of cores aren't going to be pre-filtered out anyway by the miniumum specs needed for cores that presently use native GL pipelines so this might be a compromise that benefits old devices without effecting much on newer. 

Unless it can be confirmed that those native GL calls don't glitch with this removal.

mrjschulte:
Fine with whatever you guys think is best on how to wrap these tweaks in. 

I know the midterm plan is to move Pv to a native Metal renderer at some point but these tweaks get the lower pieces of the ARM SKU's the full CRT shader in landscape mode.

JoeMatt:
Did you push changes? @mrjschulte

JoeMatt:
Closed for rebased a revised PR, #1197 1197

PR #1200, opened 2019-09-27
Update controller mappings, Poke, NES , GB ,  

Numbski:
### What does this PR do
Attempts to align PV controllers use with that of both Apple's documentation, and the original
controller layout.

I only altered the Gameboy, GBA, NES, and PokeMini .mm source files.  It's pretty straightforward.

I'm pretty sure that unless there are upstream core changes that I missed, you should be able to plug these in and build without any trouble, but then I'm an XCode/github neophyte.  Just trying to help.

https://github.com/Provenance-Emu/Provenance/issues/56

Not a fix by any stretch, but a pretty significant band-aid.  The main reason people want to re-map the buttons is because they don't align to expectations.  These changes should correct for that.

https://github.com/Provenance-Emu/Provenance/issues/56

### Questions
Just that if there's anything I'm doing wrong with this whole process, let me know.  I'm not great with compiled software, but I'll be happy to do what I can.

### DIFF
--- a/PVGB/GB/PVGBEmulatorCore.mm
+++ b/PVGB/GB/PVGBEmulatorCore.mm
@@ -259,11 +259,27 @@ const int GBMap[] = {gambatte::InputGetter::UP, gambatte::InputGetter::DOWN, gam
         (dpad.left.isPressed || pad.leftThumbstick.left.isPressed) ? gb_pad[0] |= GBMap[PVGBButtonLeft] : gb_pad[0] &= ~GBMap[PVGBButtonLeft];
         (dpad.right.isPressed || pad.leftThumbstick.right.isPressed) ? gb_pad[0] |= GBMap[PVGBButtonRight] : gb_pad[0] &= ~GBMap[PVGBButtonRight];
 
+/*
+        Begin remapping Gameboy buttons on extended controllers.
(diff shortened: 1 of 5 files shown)


JoeMatt:
Can you explain a little more what this change is changing layout wise?

@sevdestruct Has a whole spreadsheet on the button mappings.
https://docs.google.com/spreadsheets/d/1sxgeExnqDg8vdnDHyjjcDbpB0LgOnTwIOFCgaKr2ajU/edit?usp=sharing

He's probably best to review changes since he combed over them and made the original changes tfor uniformity and tested on various controllers, MFI and iCade.

does this change affect both those controller classes?

kdan84:
Not a developer or anything here, but does this take into consideration all the new controller mapping options for ios 13 &tvOS 13 made available from wwdc19?

https://developer.apple.com/videos/play/wwdc2019/616/

There should be available options now to map Start & Select & L3/R3

sevdestruct:
@kdan84 I plan on doing some updates regarding that. Thanks for the reminder. Just need some time to revisit and integrate into the mapping model.

jasarien:
Closing as it’s seen no activity for over a year. Re-open if required.

PR #1245, opened 2019-12-01
Fix a dangerous usage of inline assembly

wierton:
I ever mined usage of inline assembly in this project, and found that in file Provenance/Cores/Atari800/atari800-src/joycfg.c, line 190 and
line 214, two inline assemblies are written to try manually save and restore
registers, this may introduce some problems. To maintain the calling 
conventions of this function, none of local variables can be
declared. Save registers in a global inline assembly function and call
this function indirectly can release the limitation on local variables.

### What does this PR do
Improve the code robustness

### DIFF
--- a/Cores/Atari800/atari800-src/joycfg.c
+++ b/Cores/Atari800/atari800-src/joycfg.c
@@ -185,9 +185,16 @@ volatile int ext=0,left,key;
 volatile int hi,lo;
 volatile int buffer[100];
 
-void key_handler(void)
+void key_handler(void);
+
+asm("key_handler:"
+    "cli; pusha;"
+    "call key_handler_impl;"
+    "popa; sti;"
+    "ret;");
+
+void key_handler_impl(void)
 {
-	asm("cli; pusha");
 	raw_key = inportb(0x60);
         if (ext==2) ext=1;else /*ext 2 is used for pause*/
         if (ext==1)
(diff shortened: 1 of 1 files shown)


JoeMatt:
This is cool.  My only concern is this core is not currently a submodule. Meaning we just copy/pasted from upstream.

We can take this PR, but I also just forked atari800 into our org. Could you also make a duplicate PR in https://github.com/Provenance-Emu/atari800

This way, when I get around to making this core like some of the newers, we have our own commits that can be merged with upstream.

Since most these cores are very stable and don't see much action it's been the workflow for me to once in a while just pull from mainlain upstream onto our fork and push, so we keep our changes.


Though I suppose you could also try to PR on upstream. Or, we could PR onto our fork linked above, then PR mainline from our repo.

wierton:
> This is cool. My only concern is this core is not currently a submodule. Meaning we just copy/pasted from upstream.
> 
> We can take this PR, but I also just forked atari800 into our org. Could you also make a duplicate PR in https://github.com/Provenance-Emu/atari800
> 
> This way, when I get around to making this core like some of the newers, we have our own commits that can be merged with upstream.
> 
> Since most these cores are very stable and don't see much action it's been the workflow for me to once in a while just pull from mainlain upstream onto our fork and push, so we keep our changes.
> 
> Though I suppose you could also try to PR on upstream. Or, we could PR onto our fork linked above, then PR mainline from our repo.

Thanks for your reply, I will duplicate this pull request on https://github.com/Provenance-Emu/atari800.

PR #1261, opened 2019-12-16
[Spike] Update PVGLViewController.m to improve V-Sync on ProMotion Devices.

mrjschulte:
### What does this PR do
This will set the device's native FPS and allow for the Emu cores to self-manage their FPS' based on their self.emulatorCore.frameInterval for a smoother sync experience. [UIScreen mainScreen].maximumFramesPerSecond will list the max device FPS for 30fps on iPhone 4, 60fps for the 5> and 120fps ProMotion devices accordingly. Should also help and work on tvOS, to regulate between 50, 59.94 and 60.0, but this is currently still untested.

(Btw noticed that the self.emulatorCore.frameInterval settings for the cores should also be revisited to report a more accurate 4-6 decimal fps, at some point.) 

Check out the visual frame limiting on ProMotion devices with the Genesis Core using the scroll test on the 240p suite. The Mednafen core is not able to scale to the native Promotion FPS and is still locked to a 60fps max.

Swap between the old and new lines of code and check the Xcode GPU debug for reporting of the max FPS and check visually if the new version looks "smoother" on ProMotion devices.

Continued process to improve the V-Sync behavior in Pv.

Any thoughts @JoeMatt @jasarien @braindx ?

### DIFF
--- a/Provenance/Emulator/PVGLViewController.m
+++ b/Provenance/Emulator/PVGLViewController.m
@@ -231,12 +231,7 @@ struct RenderSettings {
 }
 
 - (void) updatePreferredFPS {
-    float preferredFPS = self.emulatorCore.frameInterval;
-    if (preferredFPS  < 10) {
-        WLOG(@"Cores frame interval (%f) too low. Setting to 60", preferredFPS);
-        preferredFPS = 30;
-    }
-
+      float preferredFPS = [UIScreen mainScreen].maximumFramesPerSecond;
     [self setPreferredFramesPerSecond:preferredFPS];
 }
 


braindx:
Ok, so let me lay out my understanding of the issues between emulation frame rate and display frame rate and let’s go from there.

As I understand it, most if not all of the systems for which Provenance is a frontend run games at a fixed frame rate based on the input ROM. For example, an NTSC game will run at 60 Hz while a PAL game will run at 50 Hz. It is not typically possible to emulate the game at a rate other than the game’s expected fixed rate without breaking other aspects of the emulation/timing.

For this reason, the ideal scenario is to have the display running in sync at the same fixed rate as the emulation core or some integer multiple of it to avoid any aliasing between the emulation rate and the display rate, i.e. a 50 Hz display rate and a 30 Hz emulation rate cannot sync up cleanly whereas a 60 Hz display rate and a 30 Hz emulation rate can.

Since many TVs support multiple display rates, Provenance would ideally change the display rate to the most ideal one supported, e.g. select a display rate that is an integer multiple of the emulation rate or, failing that, the highest rate supported.

On many devices, like iPhones and iPads, 30 Hz and 60 Hz are the only two options available which typically makes emulating things like PAL games problematic.

Ultimately, I think the best thing Provenance can probably do is:

1. Request the fixed emulation rate from the core when setting up the display rate. Provenance currently does this via self.emulatorCore.frameInterval.
2. Select the best supported display rate given the information obtained from Step 1. Provenance currently does this via GLKViewController’s setPreferredFramesPerSecond method.
3. The core should then be able to query the display rate that is ultimately selected in Step 2 and make any adjustments to its emulation, if possible. AFAIK no core actually does this.

Rather than do the above, this PR effectively modifies step 2
to always use the highest refresh rate that the display supports. This seems like a step backwards in terms of power usage and like it could cause issues where the highest display rate supported is not an integer multiple of the emulation rate, e.g. a 60 Hz NTSC title on a 144 Hz display. Additionally, I wouldn’t expect that rendering a 60 Hz emulation at 120 Hz would produce motion that is any smoother than rendering the same emulation at 60 Hz since the core wouldn’t be producing any intermediate frames.

As I mentioned at the start, this is only my cursory understanding of how everything currently works within Provenance and iOS at this time and I may be way off the mark so please correct me if I’m wrong.

JoeMatt:
I agree with @braindx , the native FPS of the core for the current mode (NTSC/PAL) should be the requested FPS.

Some cores are odd, N64 in NTSC has variable frame rate which apparently is very complex to implement. 

Some cores do support over clocking the draw rate.

If I remember correctly, cores that support odd frame rates and have them activated return that result in `self.emulatorCore.frameInterval` so it's best to leave as-is imo.

mrjschulte:
Thanks for your replies and insights guys. (Apologies for the long reply as I walk you through my process and thoughts on this.)

The core of this issue is about trying to achieve a stable V-Sync experience in Pv, and not as Brain pointed out about running a Xx Hz core on a Yy Hz capable display system which we know can't be done as-is without some kind of temporal remapping and distortion. 

The challenge I see here in Pv is that there appears to be a disconnect somewhere between what the physical devices are running their display refresh frequencies at and what we're expecting the visual result to be based on our native Core Frame intervals and GL frame buffer timings. (I also suspect the iDevices are running a somewhat dynamic refresh rate for their displays which doesn't help our cause in trying to get to the bottom of this but we'll leave that for later.)

Both of the following scenarios were confirmed and tested with the current develop/78dc628 Git branch. It would be great if you could confirm these and the following items on your end.) I've been using the 240p Test Suite Scroll with the genesis core as my base line test.

1. On a device with a known close to 60.0Hz display(59.9964x Hz actually), in my case an iPad Air Gen 1 (with everything turned off in Pv, no CRT, no multithreaded GL, etc.) I see the following type of V-Sync artifact. Using the 240P Suite scroll test the palm trees scroll by smoothly for about 10 seconds and then the scrolling becomes stutter-y for about 3.5 secs before scrolling by smoothly again for the next 10 seconds or so until it start to stutter again for about 3.5 seconds. Rinse and repeat on the clock between the 10 seconds of smooth scrolling and then the 3.x seconds of stutter-y scrolling.

2. Moving on to a native ProMotion 120Hz (reported in some display states as still only being at a 59.9756x Hz actual refresh rate) capable device such as the 10.5" 2019 iPad Pro. Here we see that the stuttering when scrolling is much more present and is not in the "then on and then off" state we see on the other 60Hz native device listed in point 1.

If we continue allow Pv to set the Genesis Core Preferred FPS to 59.922745, the reported core refresh rate, but force the GLKViewController setPreferredFramesPerSecond to the maximum value the device can support (which is 120 in the case of the above mentioned iPad Pro) we see much more stable scrolling behavior for our test case. The device's display is now over-sampling the preferred core timebase of the core's rendered frames accordingly and looking at the GPU instrument debug mode it is able to keep up. 

The stuttering at Max FPS on the ProMotion display is much reduced compared to letting it run at the Preferred Core FPS but it is still present alluding to something else altogether and perhaps out of our control in regards to how the device manages the displayed refresh rate.

With the setPreferredFramesPerSecond set to the highest device capable refresh rate it is still not as buttery smooth as what I'm seeing on a Apple TV running on a LG OLED with a known and fixed 60.0Hz refresh rate indicating that Pv can run smoothly but it feels like there might be a bigger issue we'd want to look at for the iOS devices.

Hoping to get your thoughts and insights about the above and that we can continue the relevant dialogue here in order to get to the bottom of this and improve Pv's handling of V-Sync on mobile devices.

braindx:
Ah, I see. So, as you mentioned, the easiest thing to do would be for me to boot this up and see for myself. Unfortunately, I won't have access to an Apple development machine until after the holidays.

However, what you're describing sounds like aliasing between the v-sync timing and the emulation timing. When the timing lines up well things are smooth but it sounds like eventually it doesn't render in time for the vsync and then things go south from there. There are lot of factors that probably influence why the timing is what it is, but it's unsurprising then that setting a higher framerate is smoother since missing a vsync has less of a penalty (shorter duration between frames). As you mentioned, a higher framerate doesn't solve the timing issue it just lessens the artifacts.

Vsync timing and the way the renderer synchronizes with the core depends on the advertised capabilities of the core itself. The scroll test suite you're referring to is for the Genesis core which does not render to OpenGL and does not currently support/implement double buffering. If you look at the very bottom of glkView:drawInRect: in PVGLViewController.m you can see that in this case it synchronizes with the core via @synchronized(self.emulatorCore). My best guess would be that it's that synchronization which is causing vsyncs to be missed. As a sanity check, you may want to try commenting out that line and rendering without synchronization. This will allow the render thread to render immediately at the expense of potentially rendering a frame while the Genesis core is still writing to it, which may cause tearing or other artifacts.

If removing the synchronization _does_ ameliorate the situation then we should probably look into implementing double buffering for the Genesis core as that would result in better performance and could potentially resolve the timing between CPU and GPU.

mrjschulte:
I wanted to mention that I have also been doing all of my testing in the above write up with the devices in landscape mode.

mrjschulte:
With that line, and the surrounding double buffer check blocks commented out for good measure, the scrolling performance does indeed improve for the 240P Suite Scroll test with the Genesis Core.

We can observe it for about 13 or so seconds of smooth scrolling and then a quick half of a second of stuttering before the scrolling continues for about another 13 or so seconds of smoothness before the pattern repeats itself.

mrjschulte:
(thanks for the input btw)

braindx:
I'm concerned that you're still seeing stuttering without any synchronization. Just as a sanity check: you're running Release rather than Debug correct?

It would be good to determine what's causing the stutter at this point, i.e. is it the actual emulator core that's hitching or chugging on particular frames for some reason, or one of the threads getting preempted/rescheduled by the OS, etc. Will probably just have to dig in with Instruments.

mrjschulte:
Yeah, me too.I'll do a clean pull now but I was using the Release Pv scheme for sure. 

I'm also starting to suspect that the Genesis core has timing issues at this point. I'll grab the latest RA and Genesi plus Gx and test but I know they have their own challenges with Sync on iOS.

If you do have time to dive in make sure to use the 10.3 (10G8) Xcode release to be able to use the instrument GPU debugging. GPU debugging GLES won't work on Xcode 11.

mrjschulte:
Interestingly here's what RetroArch's timing approach is:
https://github.com/libretro/docs/blob/master/docs/guides/ratecontrol.pdf

braindx:
As far as I can tell, that guide primarily discusses resampling audio while assuming video is synchronized using vsync, so that should be consistent with what we're doing here.

Section 1.2 of that document states, "As this method aims to implement a synchronization when VSync is used, this method is only useful when game frame rate is close to monitor frame rate. If this is not the case, other methods should be employed."

mrjschulte:
(Yup, Audio is another problem altogether and we'll deal with that later.)

JoeMatt:
@mrjschulte Can we make this a PR with feature flagging? 

In other words, wrap the changes in an option so we can do some baseline tests and compare them to the changes.

Instead of discussing performance we should just measure it.

(Note to self, TODO: add to Provence UI automation tests of each core with timers)

Testing is the biggest issue I have with the project right now. In my other projects, I've adopted a smaller PR, rapid development, merge everything into developing using feature flags until QA approval which heavily depends on automation tests and dependency injection using temporary class duplication rather than inline edits.

I'm going to add Swift inject soon and start making the app level classes injected so we can test new versions of view controllers, views etc without breaking current working versions, or having 'fork' branches that get so far off develop branch I never finish them.

mrjschulte:
Done some more digging and it looks like I have some more insights that I’ll add to the broken up Pr’s from this.

JoeMatt:
I'm considering this code instead, 

```swift
- (void) updatePreferredFPS {
    float frameInterval = self.emulatorCore.frameInterval;
    float maximumFPS = [UIScreen mainScreen].maximumFramesPerSecond;
    float scaler = floor(maximumFPS/frameInterval);
    
    float preferredFPS = frameInterval * scaler;
    if (preferredFPS  < 10) {
        WLOG(@"Cores frame interval (%f) too low. Setting to 60", preferredFPS);
        preferredFPS = 30;
    }

    [self setPreferredFramesPerSecond:preferredFPS];
}
```

mrjschulte:
Ok, I'm all for it. FPS in the iOS ecosystem at a HW level can have a lot of different values so let's try this and call the int 60 and see what happens.

mrjschulte:
I quickly made a test branch that implemented the suggestion above @JoeMatt and it performed visually worse than the current state when testing with the 240p Test suite on the Genesis Module.

Issue #1264, opened 2019-12-21
Update OpenVGDB to latest release. Fixes -> Artwork no longer downloading because Url has changed.


mrjschulte:
The current develop commit is trying to source the following url template for downloading the covers: https://img.gamefaqs.net/box/X/X/X/XXXXX_front.jpg

This no longer works as the host name has changed so the automatic cover downloads are no longer showing up correctly in the app.

The new URL template that appears to work is as follows:
https://gamefaqs1.cbsistatic.com/box/X/X/X/XXXXX_front.jpg

This new template has been tested with various titles that have artwork that is not showing up correctly in Pv anymore and results in the correct artwork being shown.

replaceWithSummaryOrDelete

 
### Expected: 

replaceWithExpectedResults

 
### Actual: 

replaceWithActualResults

 ### Reproduce:  

1. replaceWithSteps
2. 
3. 

# 
#### NOTES  

replaceWithNotesOrDelete

#

#### APP VERSION
- [x] 1.5 *beta*
- [ ] 1.4

#### APP SOURCE
- [x] Provenance-Emu GitHub  *&nbsp;(Official)*
- [ ] Other  *&nbsp;(Unofficial)*

#### INSTALLED BY 
- [ ] [Side-Loading](https://git.io/vpYzf)
- [x] [Building](https://git.io/vpYzv) *from Source*

#### PLATFORM
- [ ] iPhone
- [ ] iPad
- [ ] Apple TV

#### iOS/tvOS VERSION
- [x] 12.x
- [ ] 11.x
- [ ] 10.x
- [ ] 9.x
- [ ] 8.x 

-------------------------------------------
<sup> 🚫 We **DO NOT** support unofficial builds installed from 3rd-party sites. ([Official Install](https://git.io/vxABg)) </sup><br><sup> ❓ **Need help or have a suggestion?** Join our [Official Discord](https://discord.gg/4TK7PU5) </sup><br>

mrjschulte:
Looks like we need to update OPENVGDB as it shows the "URL changed" as the update feature.

https://github.com/OpenVGDB/OpenVGDB/releases/tag/v26.0

mrjschulte:
@jasarien / @JoeMatt Is this one hard to do?

mrjschulte:
Can confirm that it seems like an easy enough update by just replacing openvgdb.sqlite in Provenance/PVLibrary/PVLibrary/Resources with the latest 26.0 release.

[...]

PR #1273, opened 2020-01-11
updates openvgdb.sqlite to fix gamefaq box arts urls

antiquebeta:
### What does this PR do
Updates the openvgdb.sqlite file in PVLibrary to support the new urls for box art by gamefaq

All changes made are automatic, building with the new file doesn't need any code changes

Testing as per usual can be done

N/A

#1264


mrjschulte:
Were these cover URL’s just edited, as I think I read on #development, or did you download and update the latest db from the openvgdb site?

A quick diff vs the downloaded real updated openvgdb db would confirm but I think we’d want to stay in sync with what’s on the openvgdb as opposed to manual edits.

PR #1302, opened 2020-03-03
Set default ratio of Mupen64plus to 4:3 (AppleTV)

loicpirez:
### What does this PR do

This commit change the default aspect ratio of Mupen64plus to 4:3 (value: 1) on Apple TV.

Launch an N64 game with Mupen64plus core on Apple TV build.

On AppleTV build, the Mupen64plus default aspect ratio was 16:3 (value: 2), therefore image looked stretched.


sstgeorge:
Thanks, this is exactly what I was looking for!

PR #1328, opened 2020-05-14
Fix settings tab bar title

dnicolson:
### What does this PR do
This pull request ensures the "Settings" tab bar title appears correctly after navigating to some secondary level settings. Navigating to "Systems" or "Manage Conflicts" will change the tab bar title, it needs to be set again after navigating back.

Compare with the original [settings controller](https://github.com/Provenance-Emu/Provenance/blame/develop/ProvenanceTV/PVTVSettingsViewController.swift#L36), which seems to be unused.

Navigate to Settings, Systems and back again.

### Screenshots (if appropriate)
Before:
Screen Shot 2020-05-14 at 17 51 09

After:
Screen Shot 2020-05-14 at 17 57 53

### Questions
Secondary settings menus are somewhat inconsistent, some change the tab bar title (Systems, Manage Conflicts) and some show the view controller title (Core Options, Licenses). The view controller title is problematic it is out of the safe zone and can be truncated. Should the tab bar title be changed instead of using the view controller title?

### DIFF
--- a/Provenance/Settings/PVSettingsViewController.swift
+++ b/Provenance/Settings/PVSettingsViewController.swift
@@ -35,6 +35,7 @@ final class PVSettingsViewController: PVQuickTableViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        splitViewController?.title = "Settings"
         generateTableViewViewModels()
         tableView.reloadData()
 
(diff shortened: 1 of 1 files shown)


jasarien:
I’ll approve this PR, but to answer your question, I think that the tab bar title shouldn’t change at all when selecting items in the table.

Issue #1237, opened 2019-11-17
Fix broken CI on Azure and Bitrise


chrisballinger:
It looks like CI is broken on both Azure and Bitrise, so all pull requests will fail checks.

For example:
* https://dev.azure.com/mail0628/Provenance%20EMU/_build/results?buildId=375
   * Fails with `"##[error]TypeError: Cannot read property 'trim' of null"`
* https://app.bitrise.io/build/b52b9945e2eab00c#?tab=log
   * Fails with `Issue with input: Username, parameter not specified, authenticate-with-bitbucket-oauth@0.10.0 (exit code: 1) `

PR #1356, opened 2020-05-25
Fix Azure CI

cheif:
### What does this PR do
Fixing so that CI is run on Azure correctly.

I've also been looking into Travis, but there we have a problem with `carthage bootstrap` timing out, and I didn't look into using the `rome` caching thingy.

I'm not sure what the reason for having Azure + Travis + Bitrise is though?

### DIFF
--- a/Cores/Yabause/PVYabause.xcodeproj/project.pbxproj
+++ b/Cores/Yabause/PVYabause.xcodeproj/project.pbxproj
@@ -639,7 +639,7 @@
 			name = gen68k;
 			productName = gen68k;
 			productReference = 946C18D51714A48600C64BF9 /* gen68k */;
-			productType = "com.apple.product-type.tool";
+			productType = "com.apple.product-type.application";
 		};
 		B32775221E9C8444000FEC38 /* PVYabause-iOS */ = {
 			isa = PBXNativeTarget;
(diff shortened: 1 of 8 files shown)


jasarien:
Still failed: https://dev.azure.com/mail0628/Provenance%20EMU/_build/results?buildId=563&view=results

cheif:
Yup, but AFAIK I'll have to create a PR to get things running?

jasarien:
I’m not sure what you mean? This is a PR, and the build will run with your changes each time. So if your changes fix the issue, the build should succeed, right?

cheif:
Exactly, but I’ll need to test it in the PR, thus it’ll probably be failing for a while :)

cheif:
This is now ready to be merged, if I take a stab on Bitrise/Travis that'll be another PR, but I'm not sure we should have so many CI:s?

cheif:
Partially fixes #1237

jasarien:
Agreed. I don’t know why we even have more than one. Maybe after one didn’t work someone just added more to see if they worked!? 🤷‍♂️😂 if we have one working, stick with it, and remove the others, I guess.

jasarien:
I noticed some changes were made to project files. Does this affect compiling outside the context of CI at all?

cheif:
@jasarien Yeah, there was a problem building `gen68k` from CLI, and internet said to change it from `.tool` to `.application` and that solved it, I haven't seen anything breaking on my side, but then I'm not really sure what `gen68k` do in the first place.

Other than that it's just test related, switching to using a test plan so that all tests can be run when testing the `Provenance` target and adding some missing frameworks, basically just making sure that running tests work.

jasarien:
I don’t know for sure, but if I had to guess I’d would say that the gen68k program generates the 68000 assembly used by the emulator to do stuff. Just a guess though.

PR #1369, opened 2020-06-06
Fix some minor issues when building PVMednafen for devices

cheif:
This should make it possible to build to device again.

### DIFF
--- a/Cores/Mednafen/PVMednafen.xcodeproj/project.pbxproj
+++ b/Cores/Mednafen/PVMednafen.xcodeproj/project.pbxproj
@@ -1202,6 +1202,181 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
+		11389F98248D4F6B0089CBFF /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 089C1669FE841209C02AAC07 /* Project object */;
+			proxyType = 1;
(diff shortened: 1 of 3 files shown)


james-rantmedia:
Just had time to look at the changes for this one and it seems the main change is the removal of the SuperFX implementation from the compilation steps? Does that mean any ROMs that require SuperFX emulation won't work when using Mednafen as the SNES core?

cheif:
Exactly, but we’re already commenting out a bunch of stuff in the snes_faust core, so I’m not sure that it’ll work anyway. I’m also pretty sure that the superfx implementation is new in mednafen 1.24, so we’re not loosing anything.

jasarien:
Another thing I thought about overnight is that we're making a lot of changes to Mednafen itself to make it work within the confines of Provenance and/or Xcode... won't this make it more difficult to update Mednafen again in the future? Are the required changes being documented?

cheif:
Yeah, I created a UPDATING.md earlier with a description, but it’s not complete I’m afraid. I think a good way forward is to make sure CI builds everything, then we can catch things like this earlier at least.

jasarien:
Tested this today, it works to enable me to archive a build from scratch if I combine it with removing some of the static asserts in `MednafenGameCore.mm` I'm not sure why they still fail (all the compiler defines seem to have the correct sizes for the types, but they still fail, as they all seem to be `8`, when they're defined as `1`, `2` or `4` for their respective types, etc.). 

I'll merge this one and then add my removal of the static asserts.

Issue #1068, opened 2019-01-27
Game Library not updated on initial system import (first game per system)


sevdestruct:
Games are not showing up in library when they are the first game on a system

 
### Expected: 

Game shows after import wihtin its system section.

 
### Actual: 

Section shows but is blank until refresh.

 ### Reproduce:  

1. Import a ROM for a system that is currently empty/unintitialized.
2. Refresh Game Library in order to make it show up.

#

#### APP VERSION
- [x] 1.5 *beta*
- [ ] 1.4

#### APP SOURCE
- [x] Provenance-Emu GitHub  *&nbsp;(Official)*
- [ ] Other  *&nbsp;(Unofficial)*

#### INSTALLED BY 
- [ ] [Side-Loading](https://git.io/vpYzf)
- [x] [Building](https://git.io/vpYzv) *from Source*

#### PLATFORM
- [x] iPhone
- [x] iPad
- [x] Apple TV

#### iOS/tvOS VERSION
- [x] 12.x
- [ ] 11.x
- [ ] 10.x
- [ ] 9.x
- [ ] 8.x 

-------------------------------------------
<sup> 🚫 We **DO NOT** support unofficial builds installed from 3rd-party sites. ([Official Install](https://git.io/vxABg)) </sup><br><sup> ❓ **Need help or have a suggestion?** Join our [Official Discord](https://discord.gg/4TK7PU5) </sup><br>

PR #1371, opened 2020-06-08
Use a separate PVGameLibrary for logic stuff (part 1)

cheif:
### What does this PR do
This PR moves some stuff from VC:s (GameLibraryVC and SearchVC) into a GameLibrary struct, refactoring it to reuse code and make the VC:s more manageable.

The idea is to move even more stuff to `PVGameLibrary` so that the VC:s can focus on presenting stuff, instead of having a lot of logic inside them.

This also uses `RxRealm` and `RxDataSources` in `PVSearchVC`, to avoid having to deal with tokens etc.

Functionality should be intact, so it should work as before.

We should probably discuss if this is the correct way to go, I think so, and think going this way will help us solve a bunch of synchronization-issues, while making the code easier to reason about.
Expanding this for the whole `PVGameLibraryVC` would probably solve #1068 for instance.


jasarien:
@cheif the PR that joe merged earlier was the one that contained the commits here, right? Can we close this PR?

cheif:
It’s actually https://github.com/Provenance-Emu/Provenance/pull/1377, but we can still close this one and focus on that.

jasarien:
Oh yeah, my bad!

cheif:
Closed in favour of #1377

Issue #1344, opened 2020-05-18
"[LayoutConstraints] Unable to simultaneously satisfy constraints." warning


dnicolson:
Layout constraint warnings appear in the console.

 
### Expected: 

No warning.

 
### Actual: 

LayoutConstraints warnings.

 ### Reproduce:  

1. Ensure at least one ROM is imported
2. Run app

Tapping "Import ROMs..." also triggers a different LayoutConstraints warning.

```
[LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSAutoresizingMaskLayoutConstraint:0x6000026dc5f0 h=-&- v=-&- H:[UIView:0x7fb6db418520]-(0.4)-|   (active, names: '|':Provenance.PVGameLibraryCollectionViewCell:0x7fb6db418190 )>",
    "<NSLayoutConstraint:0x6000026d3d90 'UIIBSystemGenerated' H:[UIView:0x7fb6db418520]-(0)-|   (active, names: '|':Provenance.PVGameLibraryCollectionViewCell:0x7fb6db418190 )>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x6000026d3d90 'UIIBSystemGenerated' H:[UIView:0x7fb6db418520]-(0)-|   (active, names: '|':Provenance.PVGameLibraryCollectionViewCell:0x7fb6db418190 )>

Make a symbolic breakpoint at UIViewAlertForUnsatisfiableConstraints to catch this in the debugger.
The methods in the UIConstraintBasedLayoutDebugging category on UIView listed in <UIKitCore/UIView.h> may also be helpful.
```

#

#### APP VERSION
- [x] 1.5 *beta*
- [ ] 1.4

#### APP SOURCE
- [x] Provenance-Emu GitHub  *&nbsp;(Official)*
- [ ] Other  *&nbsp;(Unofficial)*

#### INSTALLED BY 
- [ ] [Side-Loading](https://git.io/vpYzf)
- [x] [Building](https://git.io/vpYzv) *from Source*

#### PLATFORM
- [x] iPhone
- [ ] iPad
- [ ] Apple TV

#### iOS/tvOS VERSION
- [x] 13.x
- [ ] 12.x
- [ ] 11.x
[...]

PR #1380, opened 2020-06-15
Fix NSLayoutConstraint warning

dnicolson:
### What does this PR do
This pull request prevents an NSLayoutConstraint warning from appearing in the console.

Inspect related constraints.

Test that `PVGameLibraryCollectionViewCell` games appear correctly.

https://github.com/Provenance-Emu/Provenance/issues/1344

### DIFF
--- a/Provenance/Game Library/UI/Game Library/Reusable Views/PVGameLibraryCollectionViewCell.swift
+++ b/Provenance/Game Library/UI/Game Library/Reusable Views/PVGameLibraryCollectionViewCell.swift
@@ -823,9 +823,6 @@ final class PVGameLibraryCollectionViewCell: UICollectionViewCell {
     override func awakeFromNib() {
         super.awakeFromNib()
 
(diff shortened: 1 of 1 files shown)


jasarien:
Going forward, I think it would be useful to have screenshots accompanying PRs that have changes to UI code. It would make it easier for reviewers, eg I can see this is a one-line change so it’s likely perfectly fine, but to see if the cell is still laid out correctly I’d have to do a full build.

dnicolson:
I agree that screenshots should be included. I normally do but in this instance, nothing has changed visually. I created https://github.com/Provenance-Emu/Provenance/pull/1382.

My change does fix the warning but I'm not confident on any unforeseen side-effects, maybe @JoeMatt could review as well.

cheif:
I think that this should be OK, but that we should probably remove the line above as well. IIRC there will already be constraints from `contentView` to `view`, making it already have a flexible width and height, so these lines shouldn't, have any effect (apart from adding more constraints that is).

PR #1384, opened 2020-06-16
Fix corner badge after orientation change

dnicolson:
### What does this PR do
This pull request fixes the alignment of the right corner badge after changing orientation.

Change orientation from landscape to portrait.

Before:
Screen Shot 2020-06-16 at 17 55 41

After:
Screen Shot 2020-06-16 at 17 56 13

### DIFF
--- a/Provenance/Game Library/UI/Game Library/CollectionViewController/PVGameLibraryViewController.swift
+++ b/Provenance/Game Library/UI/Game Library/CollectionViewController/PVGameLibraryViewController.swift
@@ -1056,6 +1056,8 @@ final class PVGameLibraryViewController: UIViewController, UITextFieldDelegate, 
     override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         super.viewWillTransition(to: size, with: coordinator)
 
+        collectionView?.reloadData()
+        
         transitioningToSize = size
         collectionView?.collectionViewLayout.invalidateLayout()
         if #available(iOS 10.0, *) {


Provenance/Game Library/UI/Game Library/CollectionViewController/PVGameLibraryViewController.swift
```diff
@@ -1056,6 +1056,8 @@ final class PVGameLibraryViewController: UIViewController, UITextFieldDelegate,
     override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         super.viewWillTransition(to: size, with: coordinator)
 
+        collectionView?.reloadData()
```
JoeMatt, line 1059: There’s probably a more correct fix than a brute force reload. That data hasn’t changed, but rather a side effect of a full redraw corrects incorrect coordinates in some cases.

This will be way too slow for large libraries though and will likely stutter the animation.

dnicolson:
The problem is not isolated to orientation changes, it also occurs when the view is loaded.

The `topRightBadgeTrailingConstraint` and the `topRightBadgeTopConstraint` values can be set incorrectly in the [updateImageConstraints](https://github.com/Provenance-Emu/Provenance/blob/develop/Provenance/Game%20Library/UI/Game%20Library/Reusable%20Views/PVGameLibraryCollectionViewCell.swift#L959) method due to incorrect `bounds` values in the [contentClippingRect](https://github.com/Provenance-Emu/Provenance/blob/develop/Provenance/Game%20Library/UI/Game%20Library/Reusable%20Views/PVGameLibraryCollectionViewCell.swift#L178) method.

PR #1391, opened 2020-06-22
Move most dependencies to use SPM instead of carthage

cheif:
This switches as much as currently possible to use SPM instead of carthage, the ones not switched listed below.

- HockeySDK
  - This seems to be deprecated, and we briefly talked about removing tracking altogether (or at least just use one SDK)
- XLActionController
  - This is actually imported in `PVEmulatorViewController~iOS.swift`, and a subclass is created, but it doesn't seem to be instantiated, so maybe we can just remove that code and the dependency?
- QuickTableViewController
  - Is available for installation through SPM, but forces iOS 13, I think we can just create a PR that lowers that target.
- SteamController
  - Not available for installation through SPM, maybe we can just do a PR that fixes that?
- Crashlytics/Fabric
  - Binary dependencies, thus not supported by SPM yet. Either we wait, or we stop using them (Same as HockeySDK)
- NSLogger
  - Not available for installation through SPM, and seems to be some problems providing it due to it not using ARC. Are we even using NSLogger?
- ZipArchive
  - Not available for installation through SPM, might be fixable with a PR.

## Testing this
Make sure building works for both iOS and tvOS, preferably on a clean clone. Since we still use some Carthage-stuff the `carthage bootstrap` is still needed.

## Discussion
What should we do with the dependencies not moved, some thought in the list above.


jasarien:
Thye CI failed, and the error shows that it was unable to resolve the package dependencies because of an SSH error... Maybe use HTTPS URLs for the packages in SwiftPM?

cheif:
Yeah, got to bed before having time to see CI, I’ll look into it when I get time.

jasarien:
I'd say take this opportunity to remove HockeySDK and Firebase/Crashlytics.

The rest are relatively small libraries and shouldn't cause much pain to install with Carthage for the time being.

jasarien:
I don't even know why Firebase/crashlytics was added -- it depends on the BundleID to report data to a project hosted on Firebase, and the BundleID is never the same when used by users in the wild, so it's useless.

cheif:
Unfortunately can't get this to work, getting into to much strange problems, so I'll close this PR and then maybe try again with XCode12

PR #1393, opened 2020-06-25
Fix core options crash on iPad

dnicolson:
### What does this PR do
This pull request fixes a crash on iPad when a UIAlert was invoked in core options. Alerts are now placed near the relevant table cell. Some theme code was removed to fix a height issue.

Open a UIAlert in core options with an iPad device.

After the crash fix:
PixelSnap 2020-06-25 at 20 55 12@2x

After the theme fix:
PixelSnap 2020-06-25 at 20 54 36@2x

### Questions
Does the removal of the `viewDidLayoutSubviews` override cause any issues?

### DIFF
--- a/Provenance/User Interface/Themes/UIAlertController+Theming.swift
+++ b/Provenance/User Interface/Themes/UIAlertController+Theming.swift
@@ -34,16 +34,6 @@ import UIKit
             }
         }
 
-        // view{load,willAppear,didAppear} had GFX glitches. This seems to render accuratly before animation and after
-        // Remove this method if you don't want ALL your UIAlertController's to look the same
-        open override func viewDidLayoutSubviews() {
-            super.viewDidLayoutSubviews()
-            setDefaultOverrides()
-            if let popoverPresentationController = popoverPresentationController {
-                popoverPresentationController.backgroundColor = .clear
-            }
-        }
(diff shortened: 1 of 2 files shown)


Provenance/User Interface/Themes/UIAlertController+Theming.swift
```diff
@@ -34,16 +34,6 @@ import UIKit
             }
         }
 
-        // view{load,willAppear,didAppear} had GFX glitches. This seems to render accuratly before animation and after
-        // Remove this method if you don't want ALL your UIAlertController's to look the same
-        open override func viewDidLayoutSubviews() {
-            super.viewDidLayoutSubviews()
-            setDefaultOverrides()
```
jasarien, line 41: If removing this resolves the sizing issue, it looks fine when you run it and the options still work when selected, I guess this isn’t needed.

dnicolson:
An alternative one-line fix could be to use the `alert` instead `actionSheet` here:

https://github.com/Provenance-Emu/Provenance/blob/497859bc452ccd15039a52ce6c43b93e1e139fde/Provenance/Emulator/CoreOptionsViewController.swift#L77

Screen Shot 2020-07-08 at 00 49 20

PR #1401, opened 2020-07-16
Proof of concept: Control game speed with right thumbstick

DanEEStar:
### What does this PR do

By steering the xAxis of the right thumbstick the player can control
the game speed of the emulator directly.
Press "right" -> GameSpeed.fast
Press "left" -> GameSpeed.slow
Neutral position -> GameSpeed.normal

This is basically just a proof of concept. But maybe someone has the same problems as I had...

I don't know if this makes sense to merge...

I recently started playing some older JRPGs in Provenance like Dragon Quest 1 and Final Fantasy VI. The battles especially in the beginning, can become very boring. You basically just have to press "Attack" and wait...
Also manually changing the Game Speed via the menu became boring fast...

So my idea was to map the Game Speed control to some, often unused, buttons on my XBOX-Gamepad. So I mapped the GameSpeed to the right thumbstick.

With this small change I can control the Game Speed in game directly with the controller.


james-rantmedia:
An interesting idea. Not sure if it would make sense to add to the code base in this current from though, as it may interfere with other cores that make use of the right stick In their controller layers, (PSX and N64 come to mind, initially...)

Maybe it could be morphed into a setting-controlled feature that can be turned on/off in the settings, and maybe further do it on a core by core basis?

jasarien:
Interesting idea. Not sure it should be merged in its current form though, as it might interfere with cores that use the right stick in their controller layer (PSX and n64 come to mind, initially...). 

Maybe it could be morphed into a setting-controlled feature that can be turned off and on in the settings, and maybe further do it on a core by core basis?

jasarien (changes requested):
As per my previous comment.

urherenow:
First, I'd like to say that this is sorely needed. Just from playing Metroid, and I've played it hundreds of times, I can say with 100% certainty that NES emulation is actually too fast on my iPAD and on my iPhone. Sure it may be a core issue, but this feature would certainly allow for correction.

In line with the settings suggestion above, I would like to add that it should be controlled with a slider, default emulation speed being center.

JoeMatt:
@claude can you look into this feature request and idea? It's a very old ticket, so we have a lot of new controller code and systems to deal with.

This could work, but as a user-configurable option in the pause menu or controller menus, and needs to have some support for how to work with cores with or without right thumbstick support requirements for game input.

Come up with a plan and either update this PR or make a new work queue ticket and tag this PR in it

JoeMatt:
> **Tracking issue filed:** See the detailed implementation plan posted above.
> 
> **Summary of recommendation:**
> 
> The original PR's branch is no longer in the remote, and the codebase has grown significantly (new controller systems, remapping, Delta skin support, RetroMenu overlay, etc.). Rather than trying to rebase the proof of concept, this should be implemented fresh in a new branch.
> 
> The work naturally splits into two PRs:
> 
> 1. **Settings + detection** (Tier 2, fast build) — adds the setting key, `ThumbstickSpeedControlMode` enum, and `SystemIdentifier.usesRightThumbstick`
> 2. **Input handler + UI** (Tier 6, full workspace build) — the actual thumbstick listener, OSD feedback, and settings/pause menu UI
> 
> I'll proceed with implementation once a tracking issue is created and prioritized. The plan comment above has full file-level detail.

@claude make a epic and work tickets for this work, tag and close this PR

Issue #1407, opened 2020-07-26
Broken build, imports Crashlytics, Fabric, and HockeySDK.


mrjschulte:
Looks like the current develop is somewhat broken @JoeMatt , looks like Crashlytics, Fabric, and HockeySDK are still called in the code but were purged recently in your Carthage update per https://github.com/Provenance-Emu/Provenance/commit/74e15d36a271fe60714cd4e11f2ad2a38221a722

PR #1410, opened 2020-08-07
Carthage fixes

JoeMatt:
### What does this PR do

Updates to the project file and build scripts to remove no-longer used frameworks,
1. Crashlyrics
2. Fabric
3. HockeyApp

Adds,
1. MSappCenter * frameworks

Never versions of RxSwift and sub-dependencies are failing for some people on tvOS.

https://github.com/RxSwiftCommunity/RxRealm/issues/142

#1407 

### Questions

-----
[View rendered fastlane/README.md](https://github.com/Provenance-Emu/Provenance/blob/feature/2020.08.CarthageFix/fastlane/README.md)


jasarien:
Why add app centre? It has the same issues as crashlytics - we won’t get any crash data from the wild because no one is going to install using the same bundle id that will be registered for the app on app centre.

jasarien:
If you just want to distribute through app centre, (which requires people signing up as testers, so is impractical) you don’t need the SDK integration to do that.

Issue #1345, opened 2020-05-18
Light and dark theme support


dnicolson:
Light and dark theme code exists in the project but doesn't seem to function correctly.

Code can be found in [Theme.swift](https://github.com/Provenance-Emu/Provenance/blob/develop/Provenance/User%20Interface/Themes/Theme.swift) but `darkTheme` is hardcoded in [PVAppDelegate.swift](https://github.com/Provenance-Emu/Provenance/blob/develop/Provenance/PVAppDelegate.swift#L91). The [ThemeSelectorViewController](https://github.com/Provenance-Emu/Provenance/blob/develop/Provenance/Settings/ThemeSelectorViewController.swift) class doesn't appear to be instantiated anywhere.

#

#### APP VERSION
- [x] 1.5 *beta*
- [ ] 1.4

#### APP SOURCE
- [x] Provenance-Emu GitHub  *&nbsp;(Official)*
- [ ] Other  *&nbsp;(Unofficial)*

#### INSTALLED BY 
- [ ] [Side-Loading](https://git.io/vpYzf)
- [x] [Building](https://git.io/vpYzv) *from Source*

#### PLATFORM
- [x] iPhone
- [ ] iPad
- [ ] Apple TV

#### iOS/tvOS VERSION
- [x] 13.x
- [ ] 12.x
- [ ] 11.x
- [ ] 10.x
- [ ] 9.x
- [ ] 8.x 

-------------------------------------------
<sup> 🚫 We **DO NOT** support unofficial builds installed from 3rd-party sites. ([Official Install](https://git.io/vxABg)) </sup><br><sup> ❓ **Need help or have a suggestion?** Join our [Official Discord](https://discord.gg/4TK7PU5) </sup><br>

PR #1418, opened 2020-09-05
Remove unused theme selector

dnicolson:
### What does this PR do
This pull request removes an unused view to select a light and dark theme.

https://github.com/Provenance-Emu/Provenance/issues/1345

### Questions
I experimented with conditionally removing the theming for iOS 13+ (and adding guard clauses to the Theme files):
```
if #available(iOS 13.0, *) {
    cell.textLabel?.textColor = .label
} else {
    cell.textLabel?.textColor = Theme.currentTheme.settingsCellText
}
```

But it created a lot of conditional code and design considerations.

Considering iOS 14 will be released soon, perhaps bumping the minimum version to iOS 13 and removing the theme as in https://github.com/Provenance-Emu/Provenance/pull/1265 would make sense? It would solve a lot of UI edge cases.


jasarien (approved):

jasarien:
Just to confirm, this doesn’t remove any actual theming, just the unused view to switch themes. The suggestion to move to a minimum deployment target of 13 is a bit of a leap right now. We’ll need to consider other factors before bumping any minimum versions.

PR #1423, opened 2020-09-13
Pce scanline restore

mrjschulte:
This PR restores the previously missing first 3 scanlines in the PCE module.


jasarien:
Looks like this makes a change to the mednafen source itself. Is that a good idea? It’ll make updating mednafen in the future a pain as we’ll have to reapply this patch manually each time.

mrjschulte:
This type of "scanline restore" patch has been submitted to the Mednafen developer by coders on multiple emulator frameworks/implementations that use the PCE functionality and sadly it has not been integrated by the main developer such, for one reason or another. 

In order for us to be doing the complete/correct rendering we would really need to make this change. I also don't foresee us updating Mednafen any time soon anyways so feel that we should be good with this change.

cheif:
Yeah, we still do quite a few changes to the mednafen source already. I wrote a quick note on how to upgrade, pointing to my commit from last time somewhere. We should probably link to this commit from there as well.

jasarien (approved):

PR #1444, opened 2020-11-10
Fix error message when gambatte core load state fails

sciencemanx:


sciencemanx:
@jasarien @JoeMatt: note CI failure unrelated to code change

JoeMatt (approved):

PR #1446, opened 2020-11-16
tvOS Ui Clean/Sync up (Phase 1)

mrjschulte:
### What does this PR do
This PR attempts to sync up the Ui Theming with the iOS version with the following changes:
- Consistent "dark" mode theming for tvOS, that closely matches iOS.
- Provenance now uses 7 columns to display the Games on the Library/collectionView.
- The Library/collectionView now uses higher resolution Box artwork to view on those big UHD/55"+ screens.
- The Game Box Art is now rendered with a smoother downscale filter for the artwork tiles showing less aliasing.
- Library has persistent "darkGrey" game titles, titles transition to full white when focussed.
- Section headers now show the Manufacturer and Full core names.
- Sections now have proper full width visual bottom "separators".
- Almost all of the relevant Section Cells have had their layout changed out to maximize the real estate and Ui on tvOS.
- Bunch of typo fixes and Ux clarifications with added cancel/confirm dialogs added.
- Recently Saved/Played Sections have also had some changes made to them to unify the updated theming.
- Added optimal spacing for Recently Saved/Played Sections.
- Added shadows to text overlay on some Recently Saved titles to improve legibility on bright tiles.
- Typography and layout update done on the Settings page, to aid with readability on 55"+ Screens.

Build out the relevant tvOS target for testing.


ProvenanceTV/Info.plist
```diff
@@ -73,7 +73,7 @@
 		<string>arm64</string>
 	</array>
 	<key>UIUserInterfaceStyle</key>
-	<string>Automatic</string>
+	<string>Dark</string>
```
cheif, line 76: This will force the app to Dark-mode, I don't think we'd like that?

cheif (commented):
I've tested this on my ATV, and there everything looks good.

Regarding the code, this introduces even more stuff like:
```
#if os(iOS)
   ...color = Theme.current.color...
#else 
   ...color = Color2
#endif
```
I'm wondering why we're not just using the theme-system for tvOS as well? If we'd like different colors for tvOS we could just have the same `#ifdef`:s in the theme-file, but that would be much easier to maintain in the long run I think.

Do you think that would be a better way of solving this?

mrjschulte:
Hi, thanks for taking a look! 
The proper way to theme this is to do what you suggest but this was a phase1 to get things to look similar to iOS rather quickly and I took the easier way out as I was fumbling through the code. Which leads us to the next point, this PR also includes forcing the app to be "dark" mode on tvOS as there are some issues with proper tvOS theming due to how the App was laid out with Storyboards and other items in the past. Forcing dark mode solved that. A lot of the iOS Theme conventions don't seem to carry over as nicely on tvOS when set to "light", hence to choice to go "dark" (for now).

ProvenanceTV/Info.plist
mrjschulte: See comment below for motivation.

mrjschulte:
Here's a screenshot triptych showing the current state and the reworked tvOS, with tvOS set to both Light and Dark "modes", showing the current Phase1 of the tvOS Ui Clean/Sync up.

before_after modes v02

PR #1448, opened 2020-11-24
Fix crash when refreshing library

### LINKED ISSUE #1427: Forcing the Game Library to refresh from the Settings menu will insta-minimize the app on tvOS
Forcing the Game Library to refresh from the Settings menu will insta-minimize the app on AppleTV.

Switch to the Apple TV Menu and call up the "running programs catalog view", scroll back to Provenance to continue using it.

cheif:
This fixes a crash when refreshing the library due to some thread synchronisation, the function itself is still pretty broken though, but at least this solves the crash.

I also took the liberty of removing some unused methods and debug-code.

Fixes #1427


Provenance/PVAppDelegate.swift
```diff
@@ -97,6 +97,14 @@ final class PVAppDelegate: UIResponder, UIApplicationDelegate {
...
         #endif
 
+        // Handle refreshing library
+        NotificationCenter.default.rx.notification(.PVRefreshLibrary)
+            .flatMapLatest { _ in
+                // Clear the database, then the user has to restart to re-scan
+                gameLibrary.clearLibrary()
+            }
+            .subscribe().disposed(by: disposeBag)
+
```
cheif, line 107: Having this live in AppDelegate isn't my favourite solution, but I think it's at least much better than it living in the LibraryVC as it does today, but I'm open to suggestions on moving it elsewhere.

PVLibrary/PVLibrary/RealmPlatform/RomDatabase.swift
```diff
@@ -374,19 +374,12 @@ public extension RomDatabase {
...
         try writeTransaction {
             realm.deleteAll()
         }
     }
 
-    func deleteAll<T: Object>(_ type: T.Type) throws {
-        try writeTransaction {
-            realm.delete(realm.objects(type))
-        }
-    }
```
cheif, line 388: This `deleteAll` and the one in `PVObject` wasn't used at all, so I took the liberty of removing them.

mrjschulte: Great! Thanks.

Provenance/PVAppDelegate.swift
mrjschulte: Whilst not ideal it's better here than where it was. @jasarien any other input?

jasarien (approved):

PR #1451, opened 2020-11-26
Remove some uneeded dependencies

cheif:
## What does this PR do
This PR removes `XLActionController` and `RxGesture`, reasoning below.

### XLActionController
This seems to have been used at some point, but right now it's just used for a `ViewController` that's never used, it seems like we're using a plain `UIAlertController` instead. So `XLActionController` is removed, and the dead code that was referencing it as well.

### RxGesture
`RxGesture` is only used in one single place in the app, and in that place we might as well just use a button, since we're only listening for a tap. So I changed the `UIImageView` to a `UIButton` and used the `tap` event on that instead, then I got rid of `RxGesture`

## Why did I do this
I think we have quite a lot of dependencies right now, slowing down builds and making the app bulkier. The dead code is just an annoyance, if we'd like the behaviour some time in the future we can just re-add the dependencies.


mrjschulte:
This all looks and sounds great @cheif !

jasarien (approved):

mrjschulte:
Looks like we have an issue @cheif .
> carthage bootstrap --use-ssh --platform iOS
Parse error: expected submodule commit SHA in output of task (ls-tree -z HEAD Carthage/Checkouts/RxGesture) but encountered: 

And then nothing,

jasarien:
Did you use the Carthage script for xcode 12?

mrjschulte:
I'm on Xcode 11.6. Is this a new requirement?

mrjschulte:
Thanks for the updated push @cheif !

mrjschulte (approved):

PR #1454, opened 2020-12-06
Project Release Scheme Optimization and Performance tweaks

mrjschulte:
This PR makes a few changes with the goal of increasing the Performance of the Project Release Schemes and the various builds it generates:

- reenables DoubleBuffering on the Genesis Core for smoother scrolling.
- optimizes the Release Schemes for cleaner builds with no debug executable or other profiling items left.
- per frame glClear calls are reserved for Debug builds only, helps slower hardware keep up with less GPU frame time spent on clearing the frame.
- disables nativeScale on tvOS by default, we don't need retina scaling factors applied to Core content on tvOS.


cheif (approved):
Looks like reasonably changes to me, I don't have enough context to make sense of the optimisations though 😄

Issue #1456, opened 2020-12-07
Left bar button items are inaccessible on tvOS


dnicolson:
It is not possible to switch between the filter and import buttons, accessing the left bar buttons is problematic as well. If the conflicts button is visible the import button appears to be unreachable.

 
### Expected: 

To be able to navigate to import from the filter button by pressing right.

 
### Actual: 

Pressing right shifts focus to the main menu.

 ### Reproduce:  

1. Press down, down and up
2. Press right

![Untitled](https://user-images.githubusercontent.com/2276355/101396623-e4054800-38cb-11eb-89e1-664f5d800e17.gif)

#### APP VERSION
- [x] 1.5 *beta*
- [ ] 1.4

#### APP SOURCE
- [x] Provenance-Emu GitHub  *&nbsp;(Official)*
- [ ] Other  *&nbsp;(Unofficial)*

#### INSTALLED BY 
- [ ] [Side-Loading](https://git.io/vpYzf)
- [x] [Building](https://git.io/vpYzv) *from Source*

#### PLATFORM
- [ ] iPhone
- [ ] iPad
- [x] Apple TV

#### iOS/tvOS VERSION
- [x] 14.2
- [ ] 13.x
- [ ] 12.x
- [ ] 11.x
- [ ] 10.x

-------------------------------------------
<sup> 🚫 We **DO NOT** support unofficial builds installed from 3rd-party sites. ([Official Install](https://git.io/vxABg)) </sup><br><sup> ❓ **Need help or have a suggestion?** Join our [Official Discord](https://discord.gg/4TK7PU5) </sup><br>

mrjschulte:
Seems like this is an issue that requires at least 3 titles to be present before one can navigate to them.

cheif:
I can verify this as well, I had a quick glance at the code and didn't find anything strange, and according to the documentation this should be handled well https://developer.apple.com/documentation/uikit/uinavigationitem/1624946-leftbarbuttonitems (says nothing about avoiding multiple buttons).

We could move one of the buttons to the top-right (that would make it on pair with iOS as well), but I guess the problem would still be there when we get conflicts?

mrjschulte:
Or, do we go even more extreme and remove all of the buttons and just make the Settings page the place where you do those items. It would be a shame to break that easily accessible tvOS Ux for adding titles but perhaps it's better and clearer (let alone actually usable)?

mrjschulte:
[...]

PR #1460, opened 2020-12-08
Hide collapse button on tvOS

dnicolson:
### What does this PR do
This pull request hides the collapse buttons on tvOS as they do not perform a function as they do on iOS.

Before:
Screen Shot 2020-12-08 at 19 30 33

After:
Screen Shot 2020-12-08 at 19 28 34

### Questions
Should the collapse functionality be made available on tvOS? It is actually possible for the collapse buttons to work on tvOS, it requires the following conditions/changes though:
- A section has to contain enough ROMs to be able to select the button on the right
- Duplicating the [header.collapseButton.rx.tap](https://github.com/Provenance-Emu/Provenance/blob/70f7720233561371362761f75c814a1d0806f442/Provenance/Game%20Library/UI/Game%20Library/CollectionViewController/PVGameLibraryViewController.swift#L314) functionality but replacing `tap` with `primaryAction`
- Removing the following function: https://github.com/Provenance-Emu/Provenance/blob/70f7720233561371362761f75c814a1d0806f442/Provenance/Game%20Library/UI/Game%20Library/CollectionViewController/PVGameLibraryViewController+CollectionView.swift#L84


cheif:
I totally agree with removing this, but I also think that we might want to make it work instead? We could probably modify the `collapsible` state of a section to take number of items into account, and then maybe make a convenience-thing calling tap or primaryAction depending on platform. 
There's so many #if-s in the code base already, would be nice to remove some of them, rather than adding more 😀

Anyway, great work and a really nice attention to detail!

mrjschulte:
I'd be totally up for making it work as well, that's why I put them back in there for the phase01 of the tvOS Ui updates.

dnicolson:
@mrjschulte It appears that these buttons have been the same for years?

The collapse button has no focus state, perhaps if the whole section header had one it might work. Right now the button is mostly inaccessible, if you turn on `UIFocusLoggingEnabled` you can see that every 7th item can access the button. Maybe this is something that can be fixed with https://github.com/Provenance-Emu/Provenance/issues/1456.

mrjschulte:
Yeah, noticed that the chevron has no real focus state. Ideally I think we'd like to expand the tvOS Ux to allow for the Chevron to "collapse/expand" the various System categories.

dnicolson:
I got something crude working on this [branch](https://github.com/Provenance-Emu/Provenance/compare/develop...dnicolson:section-headers-focusable?expand=1), basically the section headers are focusable and the chevron is triggered on select. I'm sure there's a better way or doing it but that's the basic concept.

mrjschulte:
Yup, it works as advertised but do you think we could limit the "focusable area" to only be the "chevron" and not the entire section header? 

Right now the entire section header "lights up" as you stated and it kinda shows some of the Ui issues (the fact that the section headers don't extend the entire width of the Screen and are gated by the global UiInsets) we still have to resolve in a Phase 2 of the tvOS Ui update.

dnicolson:
> Yup, it works as advertised but do you think we could limit the "focusable area" to only be the "chevron" and not the entire section header?
> 
I've created a new [branch](https://github.com/Provenance-Emu/Provenance/compare/develop...dnicolson:collapse-buttons-focusable?expand=1) that shifts the focus to the collapse button.

mrjschulte:
Conceptually this new branch does the right and desirable thing @dnicolson ! Thanks for taking another stab at this.

From a Ui design aspect we lost the "darkening effect" the Section header used to apply to unselected titles as the Section header is now completely transparent. The Focus on the Chevron also uses the square bounding box of the Chevron image, is there a way to have this be more elegant and only outline the actual Chevron itself? Do we need new Chevron artwork that reflects this?

Also, why was this value changed in **PVGameLibraryViewController+CollectionView.swift** on line 23?
        _#if os(tvOS)
             return 24.0		             return 40.0
         #else_
This change from 24 to 40 costs the space to place another title on the horizontal span and should be removed for a consistent Ui that maximizes the real estate on big TV's.

dnicolson:
The collapse button now animates on focus, the color also changes to white which is the same for games that are focussed. I also fixed the changes in PVGameLibraryViewController+CollectionView.swift

A focus guide needs to be added to ensure that games can be selected when pressing down to a section with less than 7 items.

mrjschulte:
Pretty great stuff with that latest update @dnicolson ! This works a treat!

mrjschulte:
Some small bits of feedback and asks:
- There would appear to be some subtle drawing/overlay items that we probably won't be able to fix but my OCD wanted to call out. When collapsing multiple systems, which is awesome we can do now btw, some of the section separators aren't drawn.
- Any chance of adding in a small 1/4 second or 1/2 second rotation animation for the Chevron when collapsing and expanding the System Title sections?
- Do we want to add support for the collapseable behavior to the "Recently Played" and "Recently Saved sections as well?

Think you can make a new PR with the branch in question so we can track these last asks more easy and remove the current "hide collapse button" PR?

All in all, really awesome work! Thanks for the contribution.

mrjschulte:
One more general wish: when Expanding and Collapsing sections it would be great if we could keep the current Section item selected. Currently, this selection appears to "reset" to the first Title in the GameViewLibrary which can make it a little disorienting to find your bearings in the Game Library when collapsing/expanding sections.

dnicolson:
> * There would appear to be some subtle drawing/overlay items that we probably won't be able to fix but my OCD wanted to call out. When collapsing multiple systems, which is awesome we can do now btw, some of the section separators aren't drawn.

I also noticed this, it appears that for every second section the bottom separator is partially hidden by the title and collapse button. This problem already exists on iOS so I will look at fixing this in a different pull request.

> * Any chance of adding in a small 1/4 second or 1/2 second rotation animation for the Chevron when collapsing and expanding the System Title sections?

> �One more general wish: when Expanding and Collapsing sections it would be great if we could keep the current Section item selected. Currently, this selection appears to "reset" to the first Title in the GameViewLibrary which can make it a little disorienting to find your bearings in the Game Library when collapsing/expanding sections.

The collapse button causes the collection views to be recreated, which makes it difficult to reuse certain objects. I was able to partially get some hack to work but it may not be a viable solution.

> * Do we want to add support for the collapseable behavior to the "Recently Played" and "Recently Saved sections as well?

Maybe it could be looked at in another pull request, iOS doesn't support this yet.

> Think you can make a new PR with the branch in question so we can track these last asks more easy and remove the current "hide collapse button" PR?

I'll clean up the branch and create a pull request.

mrjschulte:
Great, danke! Can't wait to see it.

PR #1467, opened 2020-12-20
Make collapse buttons focusable on tvOS

dnicolson:
### What does this PR do
This pull request enables the collapse buttons on tvOS, similar to how they function on iOS. It was originally planned to remove the non-functional buttons in https://github.com/Provenance-Emu/Provenance/pull/1460, after some discussion adding focus functionality was decided instead.

Look at how the focus state is changed.

Test with at least a few consoles that have ROMS with less than 7, 7, and more than 7 items.

Different focus approaches were considered:
- Originally the focus was automatically shifted from the header to the button. This unfortunately rendered some section items inaccessible as the focus would only go down from the button which required at least 7 items below for a focus hit.
- Letting the user focus on the collapse button by using the right key (not a swipe), this only worked in the Simulator and also required additional input.
- Show a focused style for the collapse button when the header is focused and trigger a tap/primaryAction event on select.

The latter focus option is used in 41668a0, it seems to be a good UX tradeoff but triggering the event perhaps is not the best solution.

gifme


mrjschulte:
Thanks for the updated PR @dnicolson ! I don't have access to my appleTV just yet so I'll have to wait with the review but as a quick glance it looks good to go and can't wait to try it out!

mrjschulte (approved):

PR #1473, opened 2021-01-01
Fix start button on tvOS

dnicolson:
### What does this PR do
This pull request fixes the menu button on an MFi controller, where the start/pause button is invoked twice on tvOS. It appears to have been erroneously introduced in #1279. A seemingly redundant code comment about unsupported tvOS 9 was removed and unused Objective-C code as well.

Check if the menu button works to invoke the alert on both iOS and tvOS only once.


mrjschulte:
Great stuff as always @dnicolson ! Will check this now.

mrjschulte:
Seems like there may still be a case this PR doesn't cover @dnicolson: 
If you disconnect a controller and then reconnect it Provenance will revert the Start Menu button behaviour this PR was to address, essentially undoing it.

dnicolson:
I also see the issue, it is not related to the code in the pull request though. The pause button is [triggered](https://github.com/dnicolson/Provenance/blob/fix-buildtime-warnings/Provenance/Emulator/PVEmulatorViewController.swift#L692) in the `controllerDidConnect` method.

It was also introduced in https://github.com/Provenance-Emu/Provenance/pull/1279/files#diff-6336dada842e7e34169c3293d64ab6aba8670b2ff6b0eff48ab6e2e041a78edcL709, when the code was only for iOS.

Testing with a Nimbus on iOS this code has no effect either way, on tvOS it causes the modal to be dismissed. Perhaps the code can be removed altogether?

mrjschulte:
Ok, thanks for the detective work. Let me try and make a quick patch build to test this out! I don't have a steam controller to test out and see if this change affects anything though.

mrjschulte:
Seems like we can take that bit out and things work as expected on tvOS. Let me put a quick patch.

mrjschulte:
This quick patch does seem to have the result that, on tvOS with MFI, the first "Menu" press after re-connecting a Controller post init of the VC isn't registered but I guess we can live with that for now.

PR #1482, opened 2021-01-03
Fix some Xcode 12 warnings

dnicolson:
### What does this PR do
This pull request resolves some warnings shown in Xcode 12. OpenGL warnings are silenced as it seems unlikely that the cores will adopt Metal. A typo was also corrected and a duplicate environment variable was removed.

Look at each commit individually.


mrjschulte:
Thanks @dnicolson ! Are you reading my mind with these fixes as I was just about to go after some warning items and reduce the number of ones we get! Let me take a look at this for iOS and tvOS and see if we can merge it in!

mrjschulte:
Do you want to also maybe take a quick stab at fixing the "Enable Base Internationalization" warnings for Mupen/Stella and PVGB whilst you're in there?

mrjschulte (approved):

PR #1489, opened 2021-01-13
Fix memory leak in pause menu

### LINKED ISSUE #1486: Provenance on iOS not releasing memory after quitting a game leading to inevitable crash!
Provenance on iOS does not release memory after quitting a game. The tvOS build does release the memory after launching a game so something odd is up between the two platforms!

----- BUG REPORT: Be clear, concise and thorough, beginning with a summary. --->
On first launch, Provenance will use around 60 or so MB of memory. Subsequently launching a game will add around 120MB to that memory footprint, bringing the new memory total to around 180MB. Quitting the game causes Provenance to still retain the same 180MB memory footprint and will not release the used memory back to the OS and lowering our memory footprint back to the original 60MB size. 

For every game launched after, the memory used by Provenance on iOS will continue to grow, even when just re-launching the same game multiple times. 

Repeatedly launching games in Provenance will cause Provenance to run out of Memory on iOS sooner or later and be killed by the OS with the following message:
_Message from debugger: Terminated due to memory issue_

 
#### APP VERSION
- [x ] 1.5 *beta* - Current develop as of Jan 9th 2021
- [ ] 1.4

#### APP SOURCE
- [x ] Provenance-Emu GitHub  *&nbsp;(Official)*
- [ ] Other  *&nbsp;(Unofficial)*

#### INSTALLED BY 
- [ ] [Side-Loading](https://git.io/vpYzf)
- [x ] [Building](https://git.io/vpYzv) *from Source*

#### PLATFORM
- [x ] iPhone
- [ ] iPad
- [ ] Apple TV

#### iOS/tvOS VERSION
- [ ] 14.x
- [x ] 13.x
- [ ] 12.x
- [ ] 11.x
- [ ] 10.x

-------------------------------------------
<sup> 🚫 We **DO NOT** support unofficial builds installed from 3rd-party sites. ([Official Install](https://git.io/vxABg)) </sup><br><sup> ❓ **Need help or have a suggestion?** Join our [Official Discord](https://discord.gg/4TK7PU5) </sup><br>

cheif:
When looking into #1486 I realised that we're running almost exactly the same code for the pause-menu for tvOS and iOS, but it's currently split into two files. So I merged them into one, keeping the important differences with compiler guards, and then fixed the memory leak in there. Also removed some uneccessary code.

### What does this PR do
Fixes a memory leak

Make sure that opening and closing games on iOS properly returns memory.

Fixes #1486


Provenance/Emulator/PVEmulatorViewController+PauseMenu.swift
```diff
@@ -201,31 +199,17 @@ extension PVEmulatorViewController {
...
-        
-        actionSheet.addAction(UIAlertAction(title: "Resume", style: .default, handler: { action in
-            actionSheet.dismiss(animated: true) {
-                DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
-                    self.core.setPauseEmulation(false)
-                    self.isShowingMenu = false
-                    self.enableControllerInput(false)
-                }
-            }
-        }))
```
cheif, line 213: This was the code responsible for the memory leak, not really sure why it was leaking (didn't really dig into it), but I guess it's either because we do the dismiss of the action-sheet (you're not supposed to do that since just selecting the action will dismiss it), or the `asyncAfter` call that somehow retains `self`.
Anyway, I found it by spotting this difference from the tvOS implementation, and made sure that everything works with this change.

Provenance/Emulator/PVEmulatorViewController+PauseMenu.swift
```diff
@@ -201,31 +199,17 @@ extension PVEmulatorViewController {
...
 
-        if actionSheet.isBeingDismissed {
-            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
-                self.core.setPauseEmulation(false)
-                self.isShowingMenu = false
-                self.enableControllerInput(false)
-            }
-        }
```
cheif, line 221: I think this maybe was some leftover from the other lib that was used? Anyway `isBeingDismissed` will always be false here, since we haven't even presented the actionSheet yet here.

mrjschulte:
Looks like this may have broken the auto save-state and screenshot capture Ux @cheif ? I'll revert and try and see what's going on.

PR #1533, opened 2021-06-23
[WIP] XCode 12.5

JoeMatt:
- Optimize Release Schemes
- make glClear calls work in Debug only
- disable nativeScale on tvOS
- Re-Enable DoubleBuffering rendering support for Genesis Core.
- Optimize Release Schemes
- Update PVSettingsModel.swift
- Mednafen - Fix Player 3 and 4 controls
- iOS XCFramework Carthage file
- Carthage tweaks for XCFramework
- Remove VERSION from Mednafen
- Remove Nix Profiling
- Xcode Project Refactor to add XCFramework support
- Provenance Scheme Tweaks


mrjschulte:
Looks like our Pv CI is going to die hard on this PR as we move to XCframeworks, unless we change it which is beyond my access and skillset. 
The Xcode project conflicts are all related to Yabause still being a thing in my Branch. Should be an easy thing to resolve.

mrjschulte:
With some quick testing on my older setup w macOS 10.14.6 and Xcode 11.7 it Looks like if we merge this any Xcode before 12 won't be able to build this app anymore.

mrjschulte:
The new Carthage build command on M1 for this Xcode 12.5 XCframeworks update would be:
`./carthage.sh bootstrap CocoaLumberjack Reachability.swift QuickTableViewController SQLite.swift SteamController RxDataSources --platform iOS,tvOS --no-use-binaries --use-xcframeworks --cache-builds & ./carthage.sh bootstrap realm-cocoa ZipArchive --platform iOS,tvOS --no-use-binaries --use-xcframeworks --cache-builds && arch -x86_64 ./carthage.sh bootstrap RxRealm --platform iOS,tvOS --no-use-binaries --use-xcframeworks --cache-builds`

You can remove the iOS/tvOS platform options accordingly to only build one of the platforms, or as needed. This Carthage command will get put into a Build Phase script for our Xcode Build targets in a future Provenance Release to make it so the users don't need to be aware of running Carthage before launching Xcode. All nice and tidy like....

Sample build times on my 8core M1 Mac mini 2021:
4min40sec for tvOS only
7min40secs for iOS only
11mins for both iOS and tvOS

JoeMatt:
> With some quick testing on my older setup w macOS 10.14.6 and Xcode 11.7 it Looks like if we merge this any Xcode before 12 won't be able to build this app anymore.

I'm ok with this. We can support older iOS's, but there's no good reason to support older XCodes.

mrjschulte:
Updated build command state:
`./carthage.sh bootstrap CocoaLumberjack Reachability.swift QuickTableViewController SQLite.swift SteamController ZipArchive --platform iOS,tvOS --no-use-binaries --use-xcframeworks --cache-builds & ./carthage.sh bootstrap realm-cocoa --platform iOS,tvOS --no-use-binaries --use-xcframeworks --cache-builds && arch -x86_64 ./carthage.sh bootstrap RxRealm --platform iOS,tvOS --no-use-binaries --use-xcframeworks --cache-builds`

I will keep this command string up to date for anyone else looking to compile and keep up with the changes as they apply to Carthage.

mrjschulte:
> > With some quick testing on my older setup w macOS 10.14.6 and Xcode 11.7 it Looks like if we merge this any Xcode before 12 won't be able to build this app anymore.
> 
> I'm ok with this. We can support older iOS's, but there's no good reason to support older XCodes.

Doesn't look like we can launch the produced app from this new Branch by itself on the previous tvOS like 13.4, it will only launch when built through Xcode directly.

JoeMatt:
> > > With some quick testing on my older setup w macOS 10.14.6 and Xcode 11.7 it Looks like if we merge this any Xcode before 12 won't be able to build this app anymore.
> > 
> > 
> > I'm ok with this. We can support older iOS's, but there's no good reason to support older XCodes.
> 
> Doesn't look like we can launch the produced app from this new Branch by itself on the previous tvOS like 13.4, it will only launch when built through Xcode directly.

It's related to realm crashes it looks like. I updated depends so there's some code changes required probaly. Looking into this.

I think we should stick with updating the dependencies, our version of RX and Realm are 18 months old at least.

xbladesub:
Installed on Apple TV week ago, not it says 'PVLibrary not found'... Any ideas why?

PR #1547, opened 2021-07-31
Mednafen: Speed up compiles

JoeMatt:
### What does this PR do

Adds a flags to a huge switch statement to not optimize that function.

Casues builds to speed up by a couple orders of magnitude. 12 seconds on my machine vs minutes just for that 1 file.


mrjschulte:
This m68k.cpp seems to be used by the Genesis (which we don't use in Pv) and Sega Saturn module in Mednafen. I'd be hesitant to get in on something as low level as this due to other unforeseen effects that could arise from this optimization.

Thoughts?

JoeMatt:
meh, this is probably stupid. closing.

PR #1553, opened 2021-08-06
Remove Carthage references and general cleanup

dnicolson:
### What does this PR do
This pull request removes Carthage references, an unused Travis CI file, and an orphaned reference to SQLite.


JoeMatt:
I looked at this. 

Some stuff I want to keep like the .gitignore's, since users updating their repo from prior versions will get a bunch of "unchecked" folders and may just confuse them. Nothing really wrong with leaving the ignores IMO.

JoeMatt (approved):

JoeMatt:
Much better, thanks @dnicolson

PR #1558, opened 2021-08-07
Update sjlj.c

mrjschulte:
Quickly hack around sigalstack to get tvOS to compile. Faust isn't our mainline SNES emu anyways but we should wrap this in a more elegant tvOS catch at a later time.


mrjschulte:
#1559

PR #1569, opened 2021-08-24
[WIP] Remove iCloud

JoeMatt:
### What does this PR do

There's a number of library rom/save files "disappearing" issues being reported. I'm pretty sure it's iCloud code being incorrectly called for non-iCloud users and/or iCloud users having issues with code we haven't tested in some time.

Since iCloud is basically not going to ever be able to provide full library sync'ing for a majority of users (free dev accounts are limited on entitlements), it's best to pull it and do something else down the road.

- Remove Entitlement
- Remove iCloud file read/write code from PVLibrary
- Remove App iCloud sync code


JoeMatt:
Closing for less complicated/destructive version #1575 1575

PR #1570, opened 2021-08-31
Fix pause-button on tvOS

cheif:
### What does this PR do
This partially reverts #1473 since that broke the pause-button behaviour (for me at least).
This PR also refactors the code related to pause/menu-button handling a bit, and uses the new `buttonMenu` API:s (if available).

I've tested on ATV 4K (tvOS 14.7) with a DS4 and the Siri remote, would be nice with some testing with other combos as well. Just verify that the pause/menu/start button brings up the provenance menu and pauses the game as expected.

It would be extra nice if @dnicolson could verify that this does indeed work, so that we're not breaking things for each other 😄


dnicolson:
Looks good to me! Tested on an Apple TV 4K running tvOS 14.7 with the new Siri Remote and a Nimbus controller.

PR #1573, opened 2021-09-03
[WIP] Add a workflow for building the app for both tvOS and iOS on github

cheif:
This builds both `iOS` and `tvOS` on PR:s, and the config is much cleaner than the azure stuff, so maybe we should consider just getting rid of azure?

NB. The iOS-check fails, due to the issue fixed in #1572


JoeMatt:
I already have a branch for this

PR #1575, opened 2021-09-05
Fix missing roms on rescan

JoeMatt:
Redo of #1569 without completely removing iCloud

- Fix some cast warnings
- gameimporter fix an unnecesary exception throw
- pblibrary re-fix the exception thing
- Fix warning in gameaudio.m
- mednafen: fix broken nes and other control inputs
- pvgamelib async call update migration
- PVFile.swift better handle missing files
- Fix iCloud or whatever is causing private in paths

### DIFF
--- a/Cores/Mednafen/MednafenGameCore.mm
+++ b/Cores/Mednafen/MednafenGameCore.mm
@@ -102,6 +102,7 @@ NSMutableDictionary *cheatList;
 int GBAMap[PVGBAButtonCount];
 int GBMap[PVGBButtonCount];
 int SNESMap[PVSNESButtonCount];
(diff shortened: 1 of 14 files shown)


mrjschulte:
I'll build this out later today.

mrjschulte:
Vanilla Build Fail
error: The package product 'NSLogger' cannot be used as a dependency of this target because it uses unsafe build flags. (in target 'PVSupport-iOS' from project 'PVSupport')

JoeMatt:
This seems to be working locally. Needs more testing, i'll break out any additional fixes into another PR so I can update and review the others first.

PR #1577, opened 2021-09-05
Fix game delete crash

dnicolson:
### What does this PR do
This pull request fixes an `Object has been deleted or invalidated` Realm crash when deleting a game.


JoeMatt:
Thanks for this. Helped me find this issue but I came up with a different solution I'm adding to another PR I'm already working on .

Rather than doing the if check here, I added it to all checks of game != using this
```swift
    public var validatedGame: PVGame? {
        return self.isInvalidated ? self : nil
    }
```

PR #1578, opened 2021-09-05
 Fix game titles if Show Game Titles is enabled

dnicolson:
### What does this PR do
This pull request sets the game title even if the Show Game Titles option is disabled. If the text is not set and Show Game Titles is enabled after adding a game the placeholder "Long Game Title: Extra Extra Long Exaggerated Long Game Subtitle" text is displayed instead.

PR #1582, opened 2021-09-06
Fix game titles if Show Game Titles is enabled

JoeMatt:
Update Provenance/Game Library/UI/Game Library/Reusable Views/PVGameLibraryCollectionViewCell.swift

Adding a proper else case to remove default string

Fix search field on iOS 11 and later


JoeMatt:
Updated version of #1578 1578

PR #1645, opened 2021-12-19
PCE/TG16 Mednafen Audio tweaks

mrjschulte:
A set of two PCE module audio related setting tweaks that enable Provenance's PCE Audio output to match much closer to the measured MDFourier output of a real system, as tested with [@artemio](https://github.com/ArtemioUrbina) from the MDFourier project.


mrjschulte:
Here is the "Before" MDFourier Plot of how the PCE/TG16 Module in Provenance looked before the tweaks:

DA__ALL_A-PCEW_72406235_F8XD-uba_vs_Mednafen_1271_SCDROM_nli_start

And below is the result of this PR as applied to the PCE/TG16 Audio Module settings with the MDFourier's Authors help:

DA__ALL_A-PCEW_72406235_F8XD-yqn_vs_MDF_Mednafen_1271_SCDteu

JoeMatt:
added to 2.0.3 in edde0f1525ab5bab350ffb7799312d393c686db4

PR #1654, opened 2021-12-23
Remove absolute path to file from .xcodeproj

davidmuzi:
Uses relative to group instead

### What does this PR do
Fixes build for tvOS

build and run on tvOS


JoeMatt:
doh! nice catch. Stupid xcode.

PR #1679, opened 2021-12-28
m68k.cpp -O0 compile time tweak

mrjschulte:
This is a quick and non code invasive Mednafen Project compilation tweak to minimize the impact of the m68k.cpp cpu module on the total Project compilation time.
 
By the way, NONE of the cores/modules we use in Mednafen appear to use the m68k cpu built by this currently.

With this simple file compilation tweak in place on a Intel 16" MacBookPro 2019, the total compilation time for a vanilla iOS Provenance Release build came down to 326.7seconds ( 5mins 27secs ) from our current vanilla build time of 616.9seconds  ( 10mins and 16secs).


JoeMatt:
this file is needed by Saturn, replacing with #1680

PR #1680, opened 2021-12-29
Speedup Mednafen compile, and LIKELY compiler hints

JoeMatt:
1. Remove m68k from hwcpu
2. Make a new libhwcpu-m68k
3. Don't compile that lib

Making a new lib in case I need to roll that back.
This should speed up release compiles a lot.

Bonus

Created LIKELY/UNLIKELY macros, similar to cores like mednafen use, to hint at how to optimize boolean logic operations and loops


pceDev16 (approved):

mrjschulte (approved):
Compiles clean in 326secs. Bunch of Mupen Controller items seem to have snuck into this PR though.

PR #1710, opened 2022-01-16
[Metal] Format and size bugfixes

ac90b671:
-Fix Red/Blue color swap when using OpenGL cores
-Avoid OpenGL init work for cores not using OpenGL


ac90b671:
Also fixed rendering when a core doesn't render to the entire CPU buffer. Added code to resize MTLTexture to screen rect of CPU buffer

ac90b671:
Also fixed format mismatch bug

JoeMatt (dismissed):

PR #1735, opened 2022-02-02
Custom UIAlertController for tvOS iOS

ToddLa:
### What does this PR do
* added custom Alert and ActionSheet class, `TVAlertController.swift`
    - on `tvOS` we use it for all Alerts and ActionSheets, on `iOS` only ActionSheets
    - will use two columns when in a compact height
    - handles being presented fullscreen or in a popover (on `iPad`)
    - top of file `TVAlertController.swift` has display and options that can be easily modified for `tvOS` and `iOS`
* added a `controllers()` function to `PVControllerManager` that should be used *instead* of directly calling `CGControllers.controllers()`
* use a modern (ie system supplied) context menu on `iOS 13+`
* fixed many bugs on `tvOS` and `iPad` where the user closes a page in a non-standard way (ie not through done button)
* fixed bug on `iPad` where navbar would disappear when search text field was given focus.
* added a keyboard based `WASD` game controller
    - useful in the `Simulator` or on an `iPad` with a keyboard
    - does not get used if an `iCade` is active
    - see below for key mappings
* made the in-game (aka pause) menu not popup from the menu/pause icon if summoned from a game controller
* made the in-game menu "driveable" with a game controller
    - added `ControllerButtonPress protocol` and `controllerUserInteractionEnabled` in `PVControllerManager` so any view in the responder chain can support this, but only in-game menu for now...

### WASD Keyboard GameController mappings:
```
[ESC:B]
[TILDE:MENU] [1:OPTIONS]
[TAB:L1]    [Q:X]     [W:UP]   [E:Y]     [R:R1]
            [A:LEFT]  [S:DOWN] [D:RIGHT] [F:B]    [RETURN:A]
[LSHIFT:L2]                              [V:R2]                     [UP]
                      [SPACE:A]                             [LEFT] [DOWN] [RIGHT]
```

all the code paths that use UIAlertController should be exercised.

The motivation for this was `tvOS` using a `UIAlertController` was a problem, many items would cause menu to scroll and would always be fullscreen.  using it on `iOS` too also made it possible to support UX navigation via game controllers.

Simulator Screen Shot - iPhone 13 mini - 2022-02-01 at 17 14 05
Simulator Screen Shot - iPhone 13 mini - 2022-02-01 at 17 13 30
Screenshot 2022-02-01 at 5 14 20 PM
Screenshot 2022-02-01 at 5 15 16 PM


JoeMatt:
merged in 3a971ec75e2f42aafd3a10f53adb6b95f57f177f

jasarien:
😍

PR #1744, opened 2022-02-05
Fix Metal input texture copy to account for cores using an offset origin

ac90b671:
Fixes CPU render and from OpenGL core paths to account for the `emulatorCore.screenRect.origin` when doing the copy into a MTLTexture.

This fixes the issue with the TurboGrafx-16 Core having the view shifted down incorrectly. 

I also briefly tested NES, N64 and Genesis cores to make sure there weren't any obvious regressions.

### DIFF
--- a/Provenance/Emulator/PVGLViewController/PVMetalViewController.m
+++ b/Provenance/Emulator/PVGLViewController/PVMetalViewController.m
@@ -543,7 +543,7 @@ PV_OBJC_DIRECT_MEMBERS
             
 
             uint outputBytesPerRow;
-            if (screenRect.origin.x == 0) // fast path if x is aligned to edge
+            if (screenRect.origin.x == 0 && (screenRect.size.width * 2 >= videoBufferSize.width)) // fast path if x is aligned to edge and excess width isn't too crazy
             {
                 outputBytesPerRow = inputBytesPerRow;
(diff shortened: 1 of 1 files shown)


ac90b671:
Also added fixes for the CRT shader sampler and an OpenGL core color channel swizzle issue on some hardware.

Provenance/Emulator/PVGLViewController/PVMetalViewController.m
```diff
@@ -325,6 +325,9 @@ - (void)updateInputTexture
     CGRect screenRect = self.emulatorCore.screenRect;
     MTLPixelFormat pixelFormat = [self getMTLPixelFormatFromGLPixelFormat:self.emulatorCore.pixelFormat type:self.emulatorCore.pixelType];
     
+    if (self.emulatorCore.rendersToOpenGL)
+        pixelFormat = MTLPixelFormatRGBA8Unorm;
+
```
JoeMatt, line 330:
```suggestion
    if (self.emulatorCore.rendersToOpenGL) {
        pixelFormat = MTLPixelFormatRGBA8Unorm;
    }    
```

JoeMatt (changes requested):

JoeMatt (approved):

PR #1748, opened 2022-02-06
Update Info.plist

mrjschulte:
Correctly set ProvenanceTV App Display name.


JoeMatt:
want to leave as `APP_DISPLAY_NAME`, that var is set in the xcconfig. If you want to change it you can edit the xcconfig instead.

PR #1987, opened 2022-10-24
Fix SwiftUI crash on iOS 16

dnicolson:
### What does this PR do
This pull request fixes a crash on iOS 16, the fix is outlined [here](https://developer.apple.com/forums/thread/709433).


JoeMatt (approved):
nice, I was digging for a while on what was causing this

PR #1988, opened 2022-10-24
General cleanup

dnicolson:
### What does this PR do
- Removes unneeded iOS conditionals
- Removes unneeded guard clause, which seems to have been left over from this [commit](https://github.com/Provenance-Emu/Provenance/commit/212f223749639a6b11a3952e7153f55a2e1020c3#diff-1ae9dfd4b77ffc8ee054e10960b9fc59512040654f1ef50a65828bd6f833fd73L680-L689).
- Removes unused variable `TEST_THEME` variable, looks to be leftover from [here]( 
https://github.com/Provenance-Emu/Provenance/commit/6e72e54d4e259b04604b38b81f464c8e2352184c#diff-241b9506fe6f4f47d23a0821a6308486ea7320e5315e47d1dbfdf7c33ca84767L45-L49).

### DIFF
--- a/PVSupport/Sources/PVSupport/Controller/HapticsManager.swift
+++ b/PVSupport/Sources/PVSupport/Controller/HapticsManager.swift
@@ -203,7 +203,7 @@ public extension HapticsManager {
 	func triple() { play(.triple) }
 }
 
-@available(iOS 13.0, tvOS 14.0, *)
+@available(tvOS 14.0, *)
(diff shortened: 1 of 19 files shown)


JoeMatt:
Looks good, though some of these conditionals may need to be modified to exclude macOS Catalyst.

I'll check that first and make any changes before merging/approving.

JoeMatt:
Replaced with #2006

PR #1997, opened 2022-11-09
Remove unused bar buttons

dnicolson:
### What does this PR do
This pull request removes non-functional tab bar buttons from the first view that appears, they also do not match the style of the main view. The `Default.xib` can actually be removed entirely without breaking anything, this does result in a black screen for a brief moment though.

The "Loading…" text has been updated for other languages as only Chinese was localized. I wonder if a more unified localization solution would be preferable as `Provenance.storyboard` doesn't appear to be localized.

Before:
Before

After:
After


JoeMatt:
Rather than remove, I updated to match the new UI in afa7e751f694d6499e9bc18277df2aec49ae7fe1

PR #1999, opened 2022-11-10
Fix tint after using the cancel action when selecting a controller

dnicolson:
### What does this PR do
This pull request prevents dismissing the modal without making a selection which leaves the cell with a tint. Dismissing the modal or selecting "Not Playing" also has the same effect.


JoeMatt (commented):
This just deletes it? 
Is there another part that does the cancel?

dnicolson:
Yeah, with this action the user can dismiss the modal by tapping outside of it.

jasarien:
What about when using a controller?

dnicolson:
The behavior does not change when using a controller.

PR #2016, opened 2022-11-17
Use TVAlertController everywhere

dnicolson:
### What does this PR do
This pull request uses `TVAlertController` for all alerts instead of ignoring `.alert` styles.

There doesn't seem to be any reason not to use this everywhere, system alerts still work and it seems to work fine with:
- Launch Web Server
- Refresh Game Library
- Empty Image Cache
- Save State Detected (a small change in 6da54ef)
- Delete Save State

Native alert

Before and after:
BeforeAfter

### DIFF
--- a/Provenance/Game Library/UI/GameLaunchingViewController.swift
+++ b/Provenance/Game Library/UI/GameLaunchingViewController.swift
@@ -637,17 +637,12 @@ extension GameLaunchingViewController where Self: UIViewController {
                     // Add a save this setting toggle
                     alert.addTextField { textField in
                         textField.text = "Auto Load Saves"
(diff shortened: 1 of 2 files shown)


JoeMatt:
Oh man, this is awesome. I never even thought of using this that way. Nice

JoeMatt (approved):

ToddLa:
Yea `TVAlertController` was designed to totally replace system one, like it does on tvOS, but when I added it I did not want to break any unknown Alert deep in a core or something....

I do remember the code that adds a progress bar needed changed on tvOS when forcing `TVAlertController` so you might want to dbl check that code path....

ToddLa:
....or maybe I was remembering the UISwitch that was added for Save States, and you already covered that. So never mind

PR #2024, opened 2022-11-25
Update PVGameLibraryViewController.swift - tvOS Header and Margin tweaks.

mrjschulte:
This (slight) tweak restores the original GameCollection Library View margins, thereby once again allowing for up to 7 titles on tvOS.  Moved the layout.sectionInsetReference addition that broke all of this on tvOS back into an already existing #if os(iOS) section.

Also reduced the Header Height extension by @dnicolson from 49px back down to 20px in order to have more real-estate for the Library view on tvOS.


dnicolson:
This leaves more space for the game library, which is good. But the active state of the buttons is very asymmetrical with the top and bottom margins.

Ideally, the vertical position of the buttons in the navigation bar buttons should be changed. So far this hasn't been so easy, maybe because it's been shoehorned in from iOS as a lot of the hacks only worked for older versions of iOS.

mrjschulte:
Yup, we all know the challenges of having a means to do a "proper" layout inside of the Navigation bar. 

Button location wise there has been some thought put into why the "add titles" button (and the conflict button if needed) are aligned in the upper left, and the other more setting oriented items are placed in the upper right alignment.

Issue #2026, opened 2022-11-27
Joystick needs fixes


JoeMatt:
1. Hide when controller not a keyboard connected
2. draggle buttons doesn't stick

Spidy123222:
Background of joystick isn't fully transparent making it possible to have a black box to cover part of screen.

PR #2030, opened 2022-11-28
Joystick fixes - draggable and clear BG, option for on/off with keyboard

JoeMatt:
refs #2026

Adds option for joystick on/off when keyboard connected

Still has some issues,

1. BG color isn't clear after drag
2. Disapears after drag


JoeMatt:
Tested last commit,
- BG is now clear, not black
- Dragging doesn't nuke the button
- Option to have visible or not when a keyboard is the main controller

PR #2042, opened 2022-12-05
Update PVMetalViewController.m

mrjschulte:
Add in  MTL Support for GL_UNSIGNED_SHORT_5_6_5 as used by SNES9X and other Cores with help from @ac90b671  .


Provenance/Emulator/PVGLViewController/PVMetalViewController.m
```diff
@@ -483,7 +483,7 @@ - (uint)getByteWidthForPixelFormat:(GLenum)pixelFormat type:(GLenum)pixelType
                 case GL_UNSIGNED_SHORT_4_4_4_4:
```
ac90b671, line 483: The `case GL_UNSIGNED_BYTE:` should still return `return 4 * typeWidth;` I think. It's just the weird formats that should return `typeWidth;` due to their size being pre-calculated

JoeMatt: noted.

i'm going to merge as-is since metal is beta and we can test and fix changes still.
@mrjschulte

PR #2048, opened 2022-12-08
Remove unneeded theming

dnicolson:
### What does this PR do
This pull request removes unneeded subclassing for styling.

The support for iOS themes on tvOS was removed as it appears to have no effect. The form controls are different and light and dark mode is hardcoded in many places on tvOS.


JoeMatt:
I need to conisder the downside of removing PV*Cell.

I kind of like having custom implementations for easy editing/overriding.

That being said, we're moving away from UIKit rapidly so maybe just accepting as-is and worrying about SwiftUI going forward is the way.

JoeMatt:
Merged all the other PRs.

This one has a small conflict now though.

PR #2065, opened 2022-12-12
Pullrequests/rf2222222/play features

JoeMatt:
Additions to #2061 

--
Added Vulkan support using MoltenMK

For now, need to install the SDK to home dir using the official installer,
https://sdk.lunarg.com/sdk/download/1.3.204.1/mac/vulkansdk-macos-1.3.204.1.dmg?Human=true

Also need to find this line and replace

```objc
   
 libPaths.push_back("@executable_path/Frameworks/libMoltenVK.dylib");
// Orig 
// libPaths.push_back("@executable_path/libMoltenVK.dylib");
```

I just need to fix the cmake script to find the framework locally. Was giving me a problem so just hacked it with "~/" install for now, since I saw that's how Play's github action for ios builds it.


JoeMatt:
merged in https://github.com/Provenance-Emu/Provenance/pull/2074

PR #2075, opened 2022-12-14
Dolphin support

JoeMatt:
Initial dolphin port. 

It builds and links, but a lot of code in the core wrapper copied from openemu needed to be commented out. 

Probably need to look instead at the dolphiniOS code rather than the copied OpenEMU stuff.

Also, BZIP2 linking isn't 100% working even though it's "in the build"

Perhaps a CMAKE flag to pick up the OS version.

See:
image


JoeMatt:
merged in ddb52b505f727ef4a2c0bf41d0af2a3e4db1d047

PR #2091, opened 2022-12-18
Fix tvOS Simulator build error

dnicolson:
### What does this PR do
This pull request fixes the `13 duplicate symbols for architecture arm64` tvOS build error in the Simulator.


JoeMatt:
thanks.
closing since i have this in #2100

PR #2092, opened 2022-12-18
General cleanup

dnicolson:
### What does this PR do
This pull request removes obsolete code, fixes deprecation warnings, and the following warning:
```
removing entry for "LSApplicationCategoryType" - only supported on macOS
```


dnicolson:
Hi @JoeMatt, it looks like only 2/6 commits got merged, is this intentional?

JoeMatt:
> Hi @JoeMatt, it looks like only 2/6 commits got merged, is this intentional?

yes, some of them broke things or weren't necessary.

Is there something obvious that i missed?

dnicolson:
The changes are minor, 4aa22ae and 28cb1f3 just remove redundant code from my testing. It's not clear what broke things or was unnecessary.

JoeMatt:
LSApplicationCategoryType should be fine, it's added by xcode. I don't think there's any bad side-effect no?

The stuff that broke was some catalyst stuff and a couple other minor things. PVSupport is needed by genesis or else archive's sometimes fail due to out of order includes.

dnicolson:
The `LSApplicationCategoryType` key was causing only warnings, I was not able to test with Catalyst. Maybe it should just be removed from the tvOS project?

I tested the PVGenesis code and it seemed fine after the orphaned `PVSupport` reference was removed in 4aa22ae. `PVSupport` is still included and works with the Xcode with the project reference `B324C63521920D6D009F4EDC`.

PR #2094, opened 2022-12-21
PVGenesis - Fix Genesis Aspect Ratio

mrjschulte:
This properly sets the Aspect Ratio for the Genesis Core in Pv.  Tested and confirmed with the 240P Monoscope pattern. This now displays a 1:1 matching Square.  Pv Integer mode now also scales accurately for the Genesis Core.

pv_ar_examples_genesis v01


JoeMatt (approved):
the floats in some of these probably isn't necessary since the struct type is float, but if it works, it works.

PR #2100, opened 2022-12-25
Single App Target

JoeMatt:
Merge iOS and tvOS app targets into one using a shared `PVUI.framework`


JoeMatt:
i manually redid the translation format fixes on develop
this branch won't merge well at all, gonna redo it

PR #2106, opened 2022-12-26
Prevent running iOS target on tvOS

dnicolson:
### What does this PR do
This pull request prevents accidentally building on a tvOS device with the iOS target.


JoeMatt:
I actually want this. I'm moving to a single app target eventually.

I had a branch with that but it got too far from develop but I'm slowly merging them into a PVApp.framework and a single app target.

Then I'm gonna clo\ne that target for a lite version with less cores for faster build testing.

dnicolson:
That sounds like a great improvement.

Would it make sense to merge at least #2091 in the interim so the build still works until it's ready?

JoeMatt:
it should already be working, testing now. I''ll just push any commits directly on develop.
thanks again!

Issue #2181, opened 2023-03-31
Provenance target can't handle spaces in source code path


smorkin:
When building the Provenance target, if the source code path has a space in it, the build will fail.

 
### Expected: 

Building should not require path without spaces.

 
### Actual: 

The script phase when building the Provenance target fails with error about no such path

 ### Reproduce:  

1. Try to build Provenance when the project path contains a space

# 
#### NOTES  

The script in the script phase for Provenance target is very simple, looking like this:
`sh ${SRCROOT}/CoresRetro/RetroArch/scripts/code-sign-cores.sh`
To fix, add "-marks (can be added around the whole path as well): 
`sh "${SRCROOT}"/CoresRetro/RetroArch/scripts/code-sign-cores.sh`

#

#### APP VERSION
Latest code

#### APP SOURCE
- Provenance-Emu GitHub  *&nbsp;(Official)*

#### INSTALLED BY 

[Building](https://wiki.provenance-emu.com/installation-and-usage/installing-provenance/building-from-source) *from Source*

#### PLATFORM
iOS

#### iOS/tvOS VERSION
16.4

-------------------------------------------
<sup> 🚫 We **DO NOT** support unofficial builds installed from 3rd-party sites. ([Official Install](https://wiki.provenance-emu.com/installation-and-usage/installing-provenance)) </sup><br><sup> ❓ **Need help or have a suggestion?** Join our [Official Discord](https://discord.gg/4TK7PU5) </sup><br>

urherenow:
If you have a fix, why don't you submit a pull request instead of a bug report? The team here are very much happy to accept even the tiniest pull requests if they are valid and don't break things. Or... did someone on the discord tell you to make this issue?

smorkin:
I asked on Discord about how to report bugs since I know some open source projects prefer to keep things a bit more internal (i.e. prefer bug reports and then fixing stuff themselves, rather than getting pull requests from outside), and was told bug reports was the way to go.

PR #2182, opened 2023-04-01
#2181: Fixed issue with spaces in source code path

smorkin:
### What does this PR do

Fixes a build script to handle spaces in the source code path

#2181


JoeMatt:
closes #2182

Issue #2167, opened 2023-02-28
Bug Report


Gandora-X:
Hi,

the most recent version 2.2.0 is crashing on Apple TV 4, tvOS 13.4.5 upon start. It has been installed via the officially named method (AltServer Sideloading).

Provenance (v. 1.5) was used on the same device with the same tvOS version and was not crashing upon launching the app.

 
### Expected: 

App would work normally since it says iOS 13 (and deriving from that tvOS 13) is supported.

 
### Actual: 

App installed successfully but crashes immediately when started.

 ### Reproduce:  

1. Sideload App version 2.2.0 on Apple TV 4 tvOS 13.4.5.
2. Launch the app.

#

#### APP VERSION
2.2.0

#### APP SOURCE
- Provenance-Emu GitHub  *&nbsp;(Official)*

#### INSTALLED BY 

-[Side-Loading](https://wiki.provenance-emu.com/installation-and-usage/installing-provenance/sideloading)

#### PLATFORM
tvOS

#### iOS/tvOS VERSION
13.4.5

JoeMatt:
Closing: a linked PR for this issue was merged. If anything remains, please open a new focused issue.

PR #2195, opened 2023-04-27
Fixed tvOS 13 and 14 crashes and also fixed and tested AirDrop support for nitoTV.

lechium:
### What does this PR do

tvOS 13 and 14 were crashing for a few different reasons, the minimum build target was missing for the fceux libraries that was causing weird missing symbol issues for libstdc++. For tvOS 13 it was crashing because of improperly gated iOS specific code in SwiftUI-Introspect. It wouldn't attempt to add or use the code but it would still create a reference to the UIColorWell symbol (which does exist in tvOS, its just unavailable). That doesn't exist at all in tvOS 13 so it would cause an exception.

for the nitoTV AirDrop code we need to use path related copying methods rather than URL because the URL based one's error out with this message:`The file couldn’t be opened because the specified URL type isn’t supported.` 

And before that error message occurs this one is printed out in os_log: `CFURLCopyResourcePropertyForKey failed because it was passed an URL which has no scheme`

this should also address the bug: https://github.com/Provenance-Emu/Provenance/issues/2167


JoeMatt:
closing in favor of #2280 since history got weird and some of the change aren't needed anymore

- adding tvos min versions is usually handeled on the global level, not sure if these should be re-added but it's been buidling without on dev
- corrected spm source change no longer required

PR #2235, opened 2023-09-26
Fix auto theme

dnicolson:
### What does this PR do
Fixes setting the theme when set to auto, previously `self.window` was not yet defined causing incorrect theming.


JoeMatt (approved):
awesome, thanks for the great work! sorry for the delay.

JoeMatt (approved):

PR #2250, opened 2023-10-31
Update RxRealm for Xcode 15

dnicolson:


JoeMatt (approved):

JoeMatt:
closing since the only diff left is a temporary .resolved reference

dnicolson:
The original commit may still be wanted as the [package directory](https://github.com/Provenance-Emu/Provenance/tree/develop/Packages/RxRealm) from #2210 seems like it might have been a temporary fix.

PR #2287, opened 2024-01-11
XCode 15 Beta and other build fixes

JoeMatt:
Fixed hardcoded XCode paths.

Reverted some accidental submodule updates to the WIP Swift PM conversion from #2155


JoeMatt:
Even though TVOS is failing, I will fix it on a minor PR where I can turn off the other platforms. 

According to others with similar error msgs, it seems to be an XCode 15.1/2 issue. There's no single solution, as it's an implicit dependency causing the problem and needs to be located and dealt with per code base.

PR #2295, opened 2024-01-28
Support tvOS in main Provenance target

JoeMatt:
A bit of a WIP but it's building  with a few caveats

- EmuThree builds but doesn't link due to included binaries only being iOS
- PVPlay same, either it built or nearly built but need tvOS vulkan
- PPSSPP same, need tvOS vulkan (we might have this already but need to convert iOS to xcframework as well)

This should allow you to select tvOS device from the primary Provenance target rather than Provenance-TV

IF this works, i'll expand to lite and xl.


urherenow:
I really wish I could give you some debug info, but I just tried to build this to my Apple TV 4k (the first one, not the current one), and it kind of bricked the TV. It's in recovery mode now. Doing a TVOS restore (I hope).

JoeMatt:
> I really wish I could give you some debug info, but I just tried to build this to my Apple TV 4k (the first one, not the current one), and it kind of bricked the TV. It's in recovery mode now. Doing a TVOS restore (I hope).

Well that's weird.

I'll try it out in a bit. My ATV installs are failing often but not specific to this branch.

urherenow:
Welp. I didn't actually change branches first. I just pulled the PR into develop. Still haven't been able to recover the TV.

JoeMatt:
> Welp. I didn't actually change branches first. I just pulled the PR into develop. Still haven't been able to recover the TV.

You're not crazy. 

I just updated this branch and finally got it to copy to my ATV 4K and instant lockup.

I was able to reboot the Apple TV though by holding the menu and home button for 15 seconds and selecting reboot from the recovery menu.

At that point, it rebooted but got stuck at boot.

Did the 15 second hold again and had to do a Reset All Settings. Rebooted after that fine.

JoeMatt:
FYI, I also had to use my iPhone as a remote over the network to setup and the remote control was not responsive either at first.

urherenow:
I tried all the resetting tips I could find online, and nothing worked. Already trashed that tv and picked up the latest gen aTV4K. Then finally figured out how to use the app signer. I can use that to sign the bot build, and it doesn't brick the tv. I'm over with trying to build it directly now. I don't want to go through that again. I even ponied up for Apple Care+ just in case the app signer method did the same thing.

PR #2303, opened 2024-02-26
Fix status bar after game quit

dnicolson:
### What does this PR do
Sets the theme again on quit to ensure the status bar color is updated.


JoeMatt (approved):
this seems like a hack for a deeper issue, but, it works.

PR #2309, opened 2024-03-03
Use eye symbol for hide item

dnicolson:
Before and after:

Screenshot 2024-03-02 at 19 02 58 Screenshot 2024-03-02 at 19 02 10


JoeMatt (approved):
nice find!

PR #2310, opened 2024-03-03
Remove deprecated 3D Touch code

dnicolson:


JoeMatt:
I always liked this feature.

What's the reason for removing other than deprecation? Does it crash or not work?

I see there's a newer API but I don't think this has been added to Provenance, has it?

https://developer.apple.com/documentation/uikit/uicontextmenuinteractiondelegate

dnicolson:
I don't recall exactly, iPhone XR was the last device to support 3D Touch and support on iOS 14 seemed inconclusive.

In any case, feel free to close this.

JoeMatt:
I made a ticket to update the APIs rather then remove.

3D touch as a hardware feature is gone but I believe it's just a rebranding for long press touch now. I don't want to lose that context menu.

https://github.com/Provenance-Emu/Provenance/issues/2318

PR #2311, opened 2024-03-03
Prevent library from shuffling when starting a game

dnicolson:
https://github.com/Provenance-Emu/Provenance/assets/2276355/6655c90c-3e76-4e3e-bf06-6619ee3eb3db


JoeMatt (approved):
this is a good fix. We have to check that recent games is updated at game quit still though

PR #2328, opened 2024-04-28
Fix incorrect BIOS file info (sizes and hashes)

MMP0:
- **Fix incorrect sizes**
Some sizes are `0` or other incorrect values.
- **Fix invalid hashes**
Some hash strings contain spaces or extra digits.


JoeMatt (approved):
Thanks! I'm going to trust these are better than the values I scraped off some wiki entries.

Will really help reduce bug reports on app store release too!

PR #2336, opened 2024-05-05
Fix Metal typo

dnicolson:
This removes both the "tm" and "™" symbol. Apple uses just "Metal" in the [developer documentation](https://developer.apple.com/documentation/metal).


JoeMatt:
Fixed in development with other settings changes; thanks for reporting.

PR #2350, opened 2024-10-11
Develop spm 2024

JoeMatt:
### What does this PR do

A lot


JoeMatt:
I think all the cores are working now, plus a bunch that weren't before, like mGBA (though it has no audio at the moment)

@rf2222222 only thing that broke is 3DS. Some roms crash, others just black screen. I updated moltenvk but doesn't seem to make a difference. Not sure what's going on and all the upstream hid their code now. the core may be dead now.

PR #2359, opened 2024-11-17
Game Importer System Logic Update

proskd:
### What does this PR do
updates game importer system determination logic.  

1.  if any method returns an exact match, return it.  
2. if none of them do, return the union of them (hopefully deduping)


JoeMatt:
merged on develop

PR #2365, opened 2024-12-23
Fixes delete roms and a few other various menu actions

proskd:
Any action relying on PVEmulatorCongifuration returning the correct path for a game would fail, because it was returning the wrong path.  We were computing the directory for the ROM incorrectly, so it was never being found.

### What does this PR do

Fixes the computation of the path for a give Game.  actions relying on that such as Delete and Move ROM should work properly now.  

Test Delete Rom, move rom, etc.


JoeMatt:
rejecting because I had the same change but slightly different using `Paths` enum instead of PVEmulatorConfiguration

PR #2368, opened 2024-12-31
[Core] Beetle PSX

JoeMatt:
### What does this PR do

- Adds a native core for Beetle PSX 

### Status

- Builds with vulkan and opengl but video doesn't render, audio starts to render until semaphore locks


JoeMatt:
no longer needed with new retroarch wrapper

PR #2371, opened 2024-12-31
[Core] Desmume2015 [WIP] 

JoeMatt:
### What does this PR do

Add Desmume2015 non-retroarch core for lite and regular.

Adds touch / mouse inputs to libretro.h wrapper

### Statsu

- Games load, video is running, and can move screens layout around
- Buttons work
- Touch DOES NOT work but touch events are forwarding to the libretro.h wrapper class


JoeMatt:
no longer needed with new retroarch wrapper

PR #3110, opened 2026-03-13
fix(mupen): restore rumble — remove stray register(nil) call + add Taptic Engine fallback

### LINKED ISSUE #2922: mupen rumble still not working
#695 part of this epic

we did a bunch of PVMupen code changes for rumble, but testing with goldeneye with a rumble pak on p1, iphone doesn't rumble

works in RetroArch mupen

JoeMatt:
## Summary

- **Bug 1 — stray `register(nil)` clears controller**: `PVMupenBridgeRumbleHelper` called `GCControllerHapticsManager.shared.register(controller: nil, forPlayer:)` on *every* rumble event. `register(nil)` removes the player's controller, so the immediately-following `rumble()` always found no controller. Fixed by removing the redundant call — `PVEmulatorCore` already registers controllers when assigned to player slots.

- **Bug 2 — no Taptic Engine fallback**: When no physical GCController is connected (iPhone with on-screen controls), `GCControllerHapticsManager.rumble()` returned silently. The `UIImpactFeedbackGenerator` fallback in `PVMupenBridgeRumbleHelper` was dead code behind an `#available(iOS 14)` guard that is always true. Fixed by adding the fallback inside `GCControllerHapticsManager.rumble()` directly.

## Files

| File | Change |
|------|--------|
| `PVMupenBridgeRumbleHelper.swift` | Remove `register()` call before `rumble()` |
| `GCControllerHapticsManager.swift` | Add `UIImpactFeedbackGenerator` fallback when no controller registered |

## Test plan
- [ ] N64 game with Rumble Pak (GoldenEye, Star Fox 64): controller motors fire on rumble events
- [ ] Same game with no physical controller (on-screen controls): iPhone Taptic Engine fires
- [ ] Haptic intensity at 0%: no rumble in either path

Closes #2922

### DIFF
--- a/.changelog/3110.md
+++ b/.changelog/3110.md
@@ -1,0 +1,6 @@
+### Fixed
+- **Mupen64Plus rumble** — removed stray `register(nil)` call that cleared the controller registration on every rumble event, causing haptics to silently fail.
+- **Rumble fallback** — added Taptic Engine fallback so on-screen/touchscreen players also feel rumble when no physical controller is connected.
+
+### Added
+- **Smart Pak (Memory + Rumble)** — new virtual combo pak mode (`PLUGIN_RAW`) for PVMupen that handles *both* persistent memory-pak saves and rumble feedback simultaneously. Previously users had to choose one or the other. Enable it by setting Controller Pak to "Smart Pak (Memory + Rumble)" in core settings. Saves are stored as `<romName>_controller<N>.mpk` in the battery-saves directory.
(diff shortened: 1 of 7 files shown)


JoeMatt:
@claude there's something i want to look into additional. The reason for this fix is we're trying to figure out a bug on why the PVRetroArch framework wrapped version of mupen rumble works automatically, even when that core automatically sets the memork pak as the p1 controller extension.

How do they do that? Can we fix our PVMupen core to work with a memory card virtuall in but also the rumble pak just works on rumble games?

also, how well does retroarch map rumble to proper haptics, meaning like different intenstity or patterns to more closely match the feel of the original hardware.

Can we do that too somehow?

JoeMatt:
@claude i merged this, but according to the last comments it sounds like you should make some follow up tickets? make sure everything on github related to this issue, epics, tickets, projects, tags, PR tags are all updated

PR #3118, opened 2026-03-13
fix(scanner): wire dynamic libretro core scanner into app init path

JoeMatt:
## Summary

- `mergeDiscoveredLibretroCores(into:)` was only called in unit tests, never from the production app boot path — so toggling the **dynamic libretro scanner feature flag** had no visible effect (legacy RetroArch core always loaded after restart)
- Make `mergeDiscoveredLibretroCores(into:)` `public` so `PVLibrary` can access it from `PVCoreLoader`
- Call it in `GameImporter.initCorePlists()` right after `CoreLoader.getCorePlists()`, before `updateCores(fromPlists:)` — the function already guards on `isFeatureEnabled` internally so it's a no-op when disabled

## Test plan

- [ ] Toggle dynamic libretro scanner feature flag ON in Settings → app restart → verify discovered dylib cores appear in core list
- [ ] Toggle feature flag OFF → app restart → verify only bundled cores appear (no regression)
- [ ] Build succeeds: `PVCoreLoader` and `PVLibrary` both compile without errors

### DIFF
--- a/PVCoreLoader/Sources/PVCoreLoader/RetroArch/PVDynamicLibretroCoreScanner.swift
+++ b/PVCoreLoader/Sources/PVCoreLoader/RetroArch/PVDynamicLibretroCoreScanner.swift
@@ -362,7 +362,7 @@ public extension CoreLoader {
     /// - Returns: Updated array that includes a synthetic "Thin Libretro" parent
     ///            containing all newly-discovered dylib cores, or the original array
     ///            unchanged if the feature flag is disabled.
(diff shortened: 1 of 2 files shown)


JoeMatt:
@claude can you look at the copilot review comment, i'm a little confused by what it means. The also we need to make sure this is in the feature flag still, i want to be able to test this but I'm assuming this is going to have bugs so don't want to replace the current core loader yet.

PR #3161, opened 2026-03-15
[Agent] fix(skins): default to preserving transparency for button overlay images

JoeMatt:
## Summary
- Fix skin button images showing black squares instead of transparent backgrounds
- Change `rep.translucent ?? false` to `rep.translucent ?? true` at all 3 call sites where button overlay images are loaded
- Also update the `UIImage(pdfData:preserveTransparency:size:)` default parameter from `false` to `true`

Most skin JSON files do not set the `translucent` field. The previous default of `false` caused `format.opaque = true` and `UIColor.black.setFill()` in the PDF rendering pipeline, producing opaque black backgrounds behind button overlay images. Transparency is the correct default for overlay assets.

Regression introduced in ec27c9814.

## Test plan
- [ ] Load a skin that does NOT set `translucent` in its JSON — button overlays should render with transparent backgrounds (no black squares)
- [ ] Load a skin that explicitly sets `translucent: false` — should still render opaque (explicit value honored)
- [ ] Load a skin that explicitly sets `translucent: true` — should render transparent (no change in behavior)

### DIFF
--- a/PVUI/Sources/PVUIBase/SwiftUI/DeltaSkins/Models/DeltaSkin.swift
+++ b/PVUI/Sources/PVUIBase/SwiftUI/DeltaSkins/Models/DeltaSkin.swift
@@ -1272,7 +1272,7 @@ extension UIImage {
             context.cgContext.drawPDFPage(page)
         }
 
-        self.init(cgImage: image.cgImage!)
+        self.init(cgImage: image.cgImage!, scale: image.scale, orientation: image.imageOrientation)
     }
 }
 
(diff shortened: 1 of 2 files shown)


JoeMatt:
@claude is copilot high with the alpha stuff? fix it maybe i dunno

JoeMatt:
@claude does copilot make a point? fix it?

JoeMatt:
@claude please fix copilot review issues

JoeMatt:
@claude more copilot review fixes to do please

JoeMatt:
@claude copilot found one remaining possible issue in a test please fix

PR #3166, opened 2026-03-15
[Agent] fix(boot): optimize game importer initialization performance (#3015)

### LINKED ISSUE #3015: fix(boot): "Initializing game importer" boot step takes disproportionately long — investigate and fix reporting
## Bug
The new boot sequence UI (BootupViewRetroWave) shows task progress. The "Initializing game importer" step appears to take an unusually long time, suggesting either:
1. GameImporter is doing expensive work (DB scan, file system walk) synchronously during boot, or
2. The task progress reporting is not accurately reflecting what step is active (fake/stale display)

## Files to investigate
- `PVUI/Sources/PVUIBase/SwiftUI/BootupViewRetroWave.swift` — boot animation and task display
- `PVUI/Sources/PVSwiftUI/App Delegate/Bootstrap/BootstrapOrchestrator.swift` — task sequence
- `PVUI/Sources/PVUIBase/State Management/AppBootupState.swift` — state machine
- `PVLibrary/Sources/PVLibrary/Importer/Services/GameImporter/GameImporter.swift` — what happens at init?

## Fix
- Identify whether GameImporter init is slow (DB queries, file walks) and defer/optimize
- Ensure progress reporting reflects actual active task, not a stale state

JoeMatt:
## Summary
- **Adds idempotency guard to `initCorePlists()`** so the redundant second call from `initSystems()` returns instantly when the background pre-fire has already completed, eliminating duplicate plist parsing and core registration.
- **Moves per-system directory creation to a background task** — `createDefaultDirectories()` was synchronously checking/creating ~60+ system ROM folders on the boot path; now runs in a `Task.detached` since directories aren't needed before the UI appears.
- **Simplifies `updateSystemToPathMap()` from async-reduce to synchronous loop** — the original used `systems.async.reduce(into:)` which created an unnecessary actor hop per system; replaced with a simple `for` loop with pre-allocated capacity.
- **Removes dead code** — `updateromExtensionToSystemsMap()` was defined inside `initSystems()` but never called.

Fixes #3015

## Test plan
- [ ] Launch app and verify boot completes without hanging on "Initializing game importer"
- [ ] Verify ROM directories are created for all systems (check after boot completes)
- [ ] Verify importing games still works correctly after boot
- [ ] Verify system path map is populated (games can be launched)
- [ ] Test cold launch (first install) and warm launch (subsequent) paths

### DIFF
--- a/.changelog/3015.md
+++ b/.changelog/3015.md
@@ -1,2 +1,7 @@
 ### Fixed
-- **Boot step "Initializing game importer" disproportionately slow** — `CoreLoader.getCorePlists()` performed a synchronous filesystem scan of the Frameworks directory on the main actor, blocking all UI updates. It now runs in a detached task. Additionally, `updateCores` was triggering a redundant full `RomDatabase.reloadCache(force: true)` during boot that duplicated the dedicated reload at 70% progress; the redundant call is removed. Fine-grained sub-step progress labels ("Loading system definitions…" → "Initializing game importer…") are now reported so the boot progress bar visibly advances during this phase
+- **Boot step "Initializing game importer" disproportionately slow** — Multiple optimizations to the game importer initialization path:
+  1. `initCorePlists()` now has an idempotency guard so the redundant second call from `initSystems()` returns instantly when the background pre-fire has already completed.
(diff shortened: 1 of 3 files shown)


JoeMatt:
@claude please review copilot review comments and fix any issues please

PR #3168, opened 2026-03-15
[Agent] feat(UX): toast warning when device audio is muted or volume is zero

JoeMatt:
## Summary
- Adds a one-time toast warning when emulation starts if the device volume is at zero or the iOS silent mode switch may be silencing audio
- Uses `AVAudioSession.outputVolume` to detect zero volume and `Defaults[.respectMuteSwitch]` + built-in speaker route detection for silent mode
- Warning shows 2 seconds after `startEmulation()` to avoid interrupting game load, and only fires once per app session (static flag)
- iOS only (`#if os(iOS)`) -- tvOS has no silent switch

## Test plan
- [ ] Launch a game with device volume at zero -- should see "Volume is at zero" warning toast
- [ ] Launch a game with silent mode switch on and "Respect Silent Mode" enabled in Settings > Audio -- should see silent mode hint toast
- [ ] Launch a second game in the same session -- toast should NOT appear again
- [ ] Launch a game with volume up and silent mode off -- no toast should appear
- [ ] Verify tvOS builds are unaffected (code is gated behind `#if os(iOS)`)


JoeMatt:
@claude fix copilot review issues please

PR #3633, opened 2026-07-29
fix(ci): publish sideload feed to a URL that actually serves

JoeMatt:
## What

`https://provenance-emu.com/altstore-source.json` — the URL printed in every alpha release note — has been a **404**. This fixes that, and makes the alpha app actually reachable from AltStore/SideStore.

## Root causes

| # | Problem | Cause |
|---|---|---|
| 1 | Feed URL 404s | Written to the pages repo **root**. That repo is a Hugo site whose `gh-pages.yml` deploys `./public`, and Hugo only copies `static/` into `public/`. |
| 2 | Moving it alone wouldn't have worked | The feed commit was tagged `[skip ci]`. `gh-pages.yml` triggers on push to `main`, and GitHub honours skip directives on push events — the tag suppressed the very deploy that publishes the file. Last `gh-pages` run before this PR was **June 1**. |
| 3 | Live `/apps.json` served `"name": null` | Two generators fought over one file: the pages repo built `static/apps.json` from `altstore-source.json`'s `app` key, while `sideload-feed.yml` overwrote that file with a schema having no `app` key. |
| 4 | Alpha would never have offered updates | Feed published `version: "3.4.0-alpha.<date>.<sha>"`. Per the [AltStore source spec](https://faq.altstore.io/developers/make-a-source), `version` must equal `CFBundleShortVersionString` — and `build.yml` stamps a plain `3.4.0`. The suffixed string is a semver **pre-release**, which sorts *below* the installed `3.4.0`. |
| 5 | Duplicate `versions` entries | iOS and tvOS IPAs emitted under one version string (18 entries, 8 unique). |

## Changes

- Write to `static/` (+ `static/apps.json`, the URL `/altstore` and `/sidestore` deep-link to and the one existing users already have — the alpha app has to appear there or nobody sees it). Each file's `sourceURL` now matches its own URL; the live `apps.json` had been self-referencing the 404.
- Drop `[skip ci]` from the feed commit.
- Alpha versioning: `version: "3.4.0"` + `buildVersion: "<commit count>"`. Verified against the IPA that produced the current alpha release — artifact is `Provenance-iOS-v3.4.0(9974).ipa`, and `git rev-list --count 9434e8ab` = `9974`. The pretty string moves to `marketingVersion` (display-only, unconstrained).
- `target_commitish` on a rolling release is the branch name, so `SHORT_SHA` was literally `"develop"`. Resolve via `/commits/alpha`.
- Drop tvOS (AltStore/SideStore are iOS-only). tvOS IPA stays on the release for manual sideload.
- Sizes from the asset JSON — `curl -sI` without `-L` only saw GitHub's 302 and reported no content-length.
- Omit `minOSVersion` on stable. AltStore *hides* versions a device can't meet, and the back catalogue reaches to 2.0.0, which predates the iOS 17 floor; stamping today's floor on every historical entry would hide the only installable builds from exactly the old devices that need them.
- Guard every alpha lookup so a miss warns and skips the alpha entry instead of aborting the step and taking the stable feed down with it.
- Remove `ALPHA_VERSION_SUFFIX` from `build.yml` — computed in two places, never consumed. Replaced with a note on why the marketing version must stay unsuffixed.

## Verification

- Generate step extracted and executed verbatim against the live API: 8 stable entries, 1 alpha, all three files valid JSON.
- Output diffed byte-for-byte against the files seeded in the pages repo — identical, so no drift on the next CI run.
- Real `hugo --minify` build confirms all three publish at the site root.
- `$EXTRA_BUILD_SETTINGS` confirmed passed to `xcodebuild` (`build.yml:331`), so the alpha bundle ID `com.provenance-emu.provenance.alpha` is real and installs alongside stable.

Companion commit: Provenance-Emu/provenance-emu.github.io@bce1739 (retires the superseded generator; already deployed).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -168,12 +168,17 @@ jobs:
       - name: Set Alpha Build Overrides
         if: github.event_name == 'push' && github.ref == 'refs/heads/develop'
         run: |
-          SHORT_SHA="${GITHUB_SHA:0:7}"
-          ALPHA_DATE=$(date -u +"%Y%m%d")
           echo "IS_ALPHA=true" >> $GITHUB_ENV
           echo "ALPHA_BUNDLE_SUFFIX=.alpha" >> $GITHUB_ENV
           echo "ALPHA_DISPLAY_NAME_SUFFIX= Alpha" >> $GITHUB_ENV
-          echo "ALPHA_VERSION_SUFFIX=-alpha.${ALPHA_DATE}.${SHORT_SHA}" >> $GITHUB_ENV
+          # MARKETING_VERSION is deliberately NOT suffixed for alpha builds.
+          # A "3.4.0-alpha.<date>.<sha>" CFBundleShortVersionString is a semver
+          # PRE-RELEASE, which sorts BELOW the plain 3.4.0 an updating user
+          # already has installed — AltStore would stop offering alpha updates.
+          # Alpha builds are distinguished by CURRENT_PROJECT_VERSION (the
+          # commit count) instead, which sideload-feed.yml publishes as the
+          # feed's buildVersion. The pretty string is derived there, for
+          # display only, as marketingVersion.
 
       - name: Set Stable Build Overrides
         if: github.event_name != 'push' || github.ref != 'refs/heads/develop'
@@ -181,7 +186,6 @@ jobs:
           echo "IS_ALPHA=false" >> $GITHUB_ENV
           echo "ALPHA_BUNDLE_SUFFIX=" >> $GITHUB_ENV
           echo "ALPHA_DISPLAY_NAME_SUFFIX=" >> $GITHUB_ENV
-          echo "ALPHA_VERSION_SUFFIX=" >> $GITHUB_ENV
 
       # Cache RetroArch prebuilt libretro dylibs (~1.4 GB uncompressed).
       # These are downloaded from the libretro buildbot during the Xcode build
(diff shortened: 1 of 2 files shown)


JoeMatt:
Addressed in d0c845de24.

**@qodo — no retry on `gh api`:** valid, fixed. Retries 4x with linear backoff. The more important half is what happens after: it now refuses to continue on an empty list. The code this PR replaced had `|| echo "[]"`, which fails *open* — an empty array still produces structurally valid JSON, so the stores read a successful publish of an app with zero versions and empty the listing. Exiting nonzero leaves the last good feed in place. Verified all four paths by extracting the step and stubbing `gh`/`sleep`: happy path exits 0 (8 stable + 1 alpha), total failure retries 4x and exits 1, empty array exits 1, recovery on the 3rd attempt exits 0.

**@sentry — `mkdir -p` before the copies:** added, but the stated premise is wrong, so flagging it rather than silently agreeing. `static/` is not untracked-or-empty — it has ~174 tracked files including the three feed JSONs, so a clone always creates it. Kept the line anyway because a `cp` into a missing directory fails with a bare ENOENT that reads like a permissions problem.

**@copilot — SideStore instructions still point at `sidestore-source.json`:** declining. That URL now serves (that's what this PR fixes), and it is deliberately not an alias for `apps.json`: it carries `identifier: com.provenance-emu.provenance.sidestore`, distinct from `...altstore`. Pointing both stores at one identifier means a user who adds the source in both apps gets a collision. `apps.json` stays canonical for AltStore because it's the URL existing users already have installed.

---

**Scope note for @qodo and @copilot's first pass:** both reviewed `Scripts/release.sh` and `.github/workflows/testflight.yml`, which are not in this PR — `gh api repos/.../pulls/3633/files` returns only `build.yml` and `sideload-feed.yml`. Those two files are already on `develop`; the diff was computed against a stale base. Copilot's second pass ("2 out of 2 changed files") is the accurate one.

The findings were real regardless, so they're split out into #3635. Chasing the `.p8` cleanup one turned up that `resolve_asc_key` was writing the App Store Connect private key **world-readable, four times per run** — details and measurements there.

PR #3637, opened 2026-07-30
fix(ci): generate + cache the libretro cheat DB for Agent Validation

JoeMatt:
Follow-up to #3636, which landed the three submodule/toolchain fixes but not this one (it was pushed after you merged).

With those in, **package resolution fully succeeds** — the exit code moved `74` → `65`, so what remains is a genuine build error:

```
error: lstat(.../PVLookup/Sources/LibretroCheatDB/Resources/libretro_cheats.sqlite.zip):
       No such file or directory (in target 'PVLookup_LibretroCheatDB')
```

`PVLookup/Package.swift:128` declares `.copy("Resources/libretro_cheats.sqlite.zip")`, but that file is gitignored (~33 MB, `Resources/.gitignore:1`) and generated from libretro-database — so it is absent from any fresh clone. `PVLookup/Scripts/generate_cheatdb_if_needed.sh` exists for precisely this case; its header reads *"used as a pre-build / pre-test step so PVLookup tests can run on CI"*. It was simply never wired into this workflow.

## Cached, not regenerated every run

Generation costs a shallow clone of libretro-database plus a python pass, so it's cached on the hash of **our** generator (`generate_cheatdb.py`, `systems.py`) rather than upstream's contents. That's deliberate: this is a compile smoke test, so it needs a *structurally valid* zip, not a current one. Keeping the real database fresh stays `update-catalogs.yml`'s job (weekly, via PR).

- Only the `.zip` is cached — that's all `Package.swift` copies; the uncompressed 138 MB `.sqlite` is a byproduct, decompressed at runtime.
- No `if: cache-hit` guard on the generate step: the script already no-ops when a valid zip is present, so always running it self-heals a cache entry that restored empty.
- Needs only `python3` + `git`. The generator imports nothing outside the stdlib, and `update-catalogs.yml` already runs it with no `pip install`.

## Note

This also means **a fresh clone can't build PVLookup** — worth its own ticket independent of CI.

### DIFF
--- a/.github/workflows/agent-validation.yml
+++ b/.github/workflows/agent-validation.yml
@@ -169,7 +169,13 @@ jobs:
   smoke-build:
     name: Build (Provenance-CI Simulator)
     runs-on: macos-26
-    timeout-minutes: 30
+    # 30 was not enough for a COLD build and created a trap: hitting the timeout
+    # kills the job before "Post Cache DerivedData" runs, so the cache is never
+    # saved, so the next run is cold too and times out again — it can never warm
+    # itself. The macos-15 -> macos-26 move invalidated the previously warm cache,
+    # which is what exposed this. 60 buys one cold build (measured: cut off at
+    # 27.8 min still building) to seed the cache; warm runs should be far under.
+    timeout-minutes: 60
     if: |
       !contains(github.event.pull_request.labels.*.name, 'build-ipa') &&
       github.event.inputs.triggered_by != 'rebase'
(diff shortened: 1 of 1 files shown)


JoeMatt:
Addressed in 88cc2b6b69.

**@sentry / @qodo — `Scripts/systems.py` in the cache key:** you're both right, fixed. I'd mirrored `update-catalogs.yml`, which gates its cheatdb job on that file, and took it as evidence of a dependency. Checking the generator directly: `generate_cheatdb.py` never imports or invokes `systems.py` — every `systems` match in it is its own SQL table names, comments and locals. So that gate is over-cautious, not a real input, and including it would only bust the cache for nothing. Replaced with `PVLookup/Scripts/generate_cheatdb_if_needed.sh`, which *is* an input (it owns the clone URL and output path).

**@copilot — "self-heals a cache entry" is misleading:** correct, and a good catch on precision. `actions/cache` doesn't re-save on a hit, so always running the generator can't repair a bad cache entry — it only guarantees the *workspace* has a usable zip for this run. Comment reworded to say exactly that, since the original would have misled whoever touches this next.

**@qodo — missing maintainer manual-apply note:** declining, because the note would be untrue here. That rule exists because the `github-actions` bot lacks the `workflows` permission and can't push workflow changes. This PR is authored under the maintainer's own credentials, not the bot's, so no manual re-push is needed — merging applies it normally. Adding the boilerplate would tell a future reader to wait for a step that isn't required.

---

Also raised `timeout-minutes` 30 → 60 in the same commit. Worth being clear that the previous run was **cancelled, not failed** — all four earlier fixes succeeded (submodule init, cheat-DB generation in 1.1 min, DerivedData cache) and the build was still running at 27.8 min when the job hit its 30-minute cap.

That cap is self-perpetuating: being killed skips `Post Cache DerivedData`, so the cache is never saved, so the next run starts cold and times out again — it can never warm itself. The `macos-15` → `macos-26` move invalidated the previously warm cache, which is what exposed it. 60 buys one cold build to seed the cache.

PR #3638, opened 2026-07-30
fix(cores): register flycast-jitless in cores.yml so it stops being pruned

JoeMatt:
`flycast-jitless` is a **shipping core** that was missing from `cores.yml` entirely:

- `Core.plist` declares `flycast-jitless.libretro.framework` — *"Flycast JIT-less (Provenance Beta)"*, `PVAppStoreDisabled=False`
- `CoreCapabilities.json:167` has an entry for it
- `PVThinLibretroCore.swift:393`: *"flycast on iOS runs JITless (`flycast-jitless`)"*
- `SystemIdentifier.swift:387` names it as the Dreamcast core

It's the App-Store-legal Dreamcast/NAOMI core — the JIT build beside it is `PVAppStoreDisabled=True`.

## Why this was dangerous, not just untidy

`get-modules.sh`'s `prune_dylibs_not_in_manifest()` deletes any dylib in `modules/` that is neither listed in a `urls*.txt` manifest nor flagged local:

```sh
is_local_dylib "$base" && continue
if ! grep -qx "$base" "$expected"; then rm -f "$dylib"; fi
```

`flycast-jitless_libretro.dylib` is a custom build in no manifest with no `local` flag — so a 13 MB artifact that **cannot be re-downloaded from the buildbot** was one prune away from `rm -f`.

## The fix

Registering with `local: true` gives it the same protection `virtualjaguar` already has. `is_local_dylib()` reads `cores.yml`, so the guard travels with the repo — unlike a `modules/*.local` sentinel, which can't, because `modules/.gitignore` is `*` and so protects only the machine that created it.

Lists regenerated via `Scripts/generate_core_lists.py`. The semantics are worth recording:

| | url lists | xcfilelists |
|---|---|---|
| `local: true` | commented → never downloaded | — |
| `enabled: true` | — | listed → still embedded |

Which is exactly right for a hand-placed dylib. Verified in the generated output: commented at `urls*.txt:29`, present at `output_modules*.xcfilelist:29`.

## One thing to check

I set `tvos: true` to match the `flycast` entry, since tvOS has the same JIT restriction. **If you only build an iOS jitless dylib, flip it to `false`** — the tvOS xcfilelists now expect `modules/flycast-jitless_libretro.dylib` to exist.

## Unrelated inconsistency spotted

`flycast` (JIT) has `appstore: true` in `cores.yml` but `PVAppStoreDisabled=True` in `Core.plist`, which `cores.yml`'s own header says to keep in sync. Result: `urls-appstore.txt:28` and `urls-appstore-tv.txt:28` both pull the JIT dylib into App Store variants. Left alone here since fixing it changes what ships — worth a separate look.

---

> [!NOTE]
> **Low Risk**
> Manifest-only change for a local dylib; main caveat is CI must still supply `flycast-jitless_libretro.dylib` in `modules/` since it is not downloaded from the buildbot.
> 
> **Overview**
> Registers the **App Store–legal** `flycast-jitless` libretro core in `cores.yml` so generated manifests match what `Core.plist` already ships.
> 
> The new entry uses `local: true` and `flycast-jitless_libretro.dylib`, which **comments out** buildbot URLs in `urls.txt` / `urls-appstore.txt` and **adds** the dylib to `output_modules.xcfilelist` and `output_modules_appstore_ios.xcfilelist`. That embeds the hand-placed binary in iOS (and App Store iOS) builds instead of leaving it declared in the app but absent from the module lists. **`tvos: false`** avoids pulling the iOS-only jitless slice into tvOS xcfilelists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aef061bede5c1ed3db198a1d89a1818aa7e0fe7e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CoresRetro/RetroArch/scripts/cores.yml
+++ b/CoresRetro/RetroArch/scripts/cores.yml
@@ -231,18 +231,28 @@ cores:
 
   - name: flycast-jitless
     # Dreamcast / NAOMI / NAOMI 2 / Atomiswave — the App-Store-legal Flycast.
-    # iOS and tvOS do not permit WX JIT pages outside a debug entitlement, so the
-    # JIT build above is PVAppStoreDisabled and this interpreter build is what
-    # ships (Core.plist: flycast-jitless.libretro.framework, "Flycast JIT-less
+    # iOS does not permit WX JIT pages outside a debug entitlement, so the JIT
+    # build above is PVAppStoreDisabled and this interpreter build is what ships
+    # (Core.plist: flycast-jitless.libretro.framework, "Flycast JIT-less
     # (Provenance Beta)", PVAppStoreDisabled=False).
     #
-    # local: true — this is a custom Provenance build, NOT on the libretro
-    # buildbot, so it is commented out of the url lists and get-modules.sh will
-    # neither download nor prune it. The dylib must be placed in modules/ by hand.
-    # It was missing from this file entirely, which meant the prune step saw an
-    # unlisted, unprotected dylib and deleted it.
+    # local: true — a custom Provenance build, NOT on the libretro buildbot, so it
+    # is commented out of the url lists and get-modules.sh will neither download
+    # nor overwrite it. The dylib is placed in modules/ by hand (or via
+    # scripts/local-dylibs.sh). Note get-modules.sh ALSO protects it by filename
+    # via LOCAL_DYLIB_PATTERNS=("*-jitless*"), so this entry is not what saves it
+    # from the prune step — what it fixes is the core being absent from the
+    # generated xcfilelists, i.e. declared in Core.plist but never in the build
+    # manifest.
+    #
+    # tvos: false — the dylib we ship is iOS-only (otool: LC_BUILD_VERSION
+    # platform 2, arm64). `filename` is platform-neutral, so tvos: true would put
+    # that iOS binary into the tvOS xcfilelists and fail the tvOS build at link
+    # time. Flip to true only alongside a tvOS-built dylib, which needs a
+    # per-platform filename split (ios_filename / tvos_filename) since one neutral
+    # name cannot carry both slices.
     ios: true
-    tvos: true
+    tvos: false
     appstore: true
     enabled: true
     filename: "flycast-jitless_libretro.dylib"
--- a/CoresRetro/RetroArch/scripts/cores.yml
+++ b/CoresRetro/RetroArch/scripts/cores.yml
@@ -222,12 +222,18 @@ cores:
     appstore: true
     enabled: true
   - name: flycast
-    # Dreamcast
+    # Dreamcast / NAOMI / NAOMI 2 / Atomiswave — the libretro buildbot build.
+    # On iOS this is effectively JIT-only: it does ship an interpreter fallback,
+    # but it is too slow to be usable, which is the reason the flycast-jitless
+    # fork below exists. JIT needs WX pages that the App Store does not permit,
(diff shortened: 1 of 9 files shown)


JoeMatt:
Addressed in `aef061bede`. Both bots were right, and one caught a factual error in my own justification.

**@sentry — `tvos: true` with a platform-neutral filename (HIGH):** correct, fixed. I checked the actual artifact rather than reasoning about it:

```
$ otool -l modules/flycast-jitless_libretro.dylib | grep -A4 LC_BUILD_VERSION
   platform 2        # iOS
      minos 17.0
$ lipo -info ...
   Non-fat file: ... is architecture: arm64
```

So the dylib genuinely is iOS-only, and `tvos: true` would have fed an iOS binary into the tvOS xcfilelists and failed the tvOS build at link time. Set `tvos: false` — the tvOS lists no longer reference it, the iOS ones still do. Enabling tvOS later needs an `ios_filename` / `tvos_filename` split, since one neutral filename can't carry both platform slices.

**@copilot — the prune rationale was wrong:** you're right, and this was the more valuable catch because it was my claim, not the code's fault. `get-modules.sh:14` defines `LOCAL_DYLIB_PATTERNS=( "*-jitless*" )` and `is_local_dylib()` tests it *first*:

```sh
for pat in "${LOCAL_DYLIB_PATTERNS[@]}"; do
    case "$base" in ${pat}*) return 0 ;; esac
done
```

So flycast-jitless was already protected by filename regardless of `cores.yml`. My "one prune away from `rm -f`" framing was false, and the commit message and comment both said so. Rewritten to state what this entry actually fixes: the core is declared in `Core.plist` and `CoreCapabilities.json` but was missing from the build manifest, so it appeared in **no generated xcfilelist**. Still worth doing — just not for the reason I gave.

**@copilot — `flycast` `appstore: true` vs `PVAppStoreDisabled=true`:** confirmed real, deliberately not fixed here. Correcting it removes a core from App Store builds, which is a shipping decision rather than a cleanup, and it predates this PR. Flagged for @JoeMatt to call — the mismatch means `urls-appstore*.txt:28` currently pulls the JIT dylib into App Store variants.

JoeMatt:
bugbot run

JoeMatt:
Added in `739be9b038` — your clarification made this unambiguous rather than a judgement call.

Since the buildbot flycast is JIT-only in practice on iOS (the interpreter fallback exists but is too slow to ship, which is why the optimised fork was made), `PVAppStoreDisabled=true` in `Core.plist` is correct and `cores.yml`'s `appstore: true` was the bug. Now follows the same pattern `dolphin` already uses — `appstore: false` plus an `appstore_excluded_reason`.

After regenerating:

| Variant | Dreamcast core |
|---|---|
| Sideload iOS | flycast (JIT) + flycast-jitless |
| **App Store iOS** | flycast-jitless only ✅ |
| Sideload tvOS | flycast (JIT) |
| **App Store tvOS** | **none** ⚠️ |

## Please read the tvOS row before merging

App Store tvOS now ships **no Dreamcast core at all**. It previously had the JIT build — which shouldn't have been there — but the replacement is iOS-only (`LC_BUILD_VERSION platform 2`), so removing the JIT one leaves nothing behind it.

Fixing that needs an `ios_filename` / `tvos_filename` split on the `flycast-jitless` entry plus a tvOS-built dylib in the store. Flagging rather than papering over it: tvOS App Store hasn't launched yet, so this may be fine to land as-is and close the gap before submission — but it should be a decision, not a surprise.

PR #3640, opened 2026-08-02
feat(cores): store custom-built dylibs in a private release store

JoeMatt:
`flycast-jitless` and `virtualjaguar` are custom Provenance builds, not on the libretro buildbot. `cores.yml` marks them `local: true`, which stops `get-modules.sh` downloading or overwriting them — but says nothing about **where they live**.

Since `modules/.gitignore` is `*`, they existed only on the machine that built them: no way to move to a new checkout, no way for CI to obtain them, and one `rm -rf modules/` from unrecoverable.

## Approach

`local-dylibs.sh` with `upload` / `fetch` / `list`, backed by release assets on the private **`Provenance-Emu/provenance-dylibs`**.

Release assets rather than committing: these are 2–13 MB binaries that get rebuilt regularly, which doesn't belong in git history. Private because they're unreleased custom builds.

```bash
CoresRetro/RetroArch/scripts/local-dylibs.sh upload   # after building locally
CoresRetro/RetroArch/scripts/local-dylibs.sh fetch    # fresh clone / CI
CoresRetro/RetroArch/scripts/local-dylibs.sh list     # expected vs on disk
```

## Design notes

- **Reads `cores.yml` for `local: true`**, so the store follows the manifest automatically instead of duplicating a list that would drift out of sync.
- **Handles both dylib shapes**: a platform-neutral `filename` (flycast-jitless) and the `_ios`/`_tvos` pair (virtualjaguar).
- **Zips with `zip -j` from inside `modules/`**, so archives contain a bare dylib matching the buildbot layout `get-modules.sh` already unzips.
- **Parses YAML with awk, not PyYAML** — this has to run on a bare runner and a fresh clone, where a pip dependency is exactly what isn't available.
- **`fetch` exits nonzero on a missing asset.** These dylibs are in the xcfilelists, so a silent miss just fails the build later with a vaguer error.

## Verification

Round-trip tested against the real store: uploaded both virtualjaguar dylibs, deleted one from `modules/`, ran `fetch`, and confirmed the restored files are **byte-identical** by sha256. The missing-asset path was verified to exit `1` (my first check of that was masked by a pipe — the exit code was `tail`'s, not the script's).

## Open question

CI can't use this yet: `GITHUB_TOKEN` can't read another private repo, so a workflow step would need a PAT with `repo` scope exported as `GH_TOKEN`. Worth deciding whether CI should pull these at all, or whether this stays a local-developer tool — the `Provenance-CI` smoke target doesn't build emulator cores, so it may not need them.

## Note

`flycast-jitless` isn't in the store yet — `cores.yml` only learns about it in #3638. Re-running `upload` after that merges will include it.

---

> [!NOTE]
> **Low Risk**
> Build-tooling only; no app runtime, auth, or data-path changes. CI still needs a PAT if workflows call `fetch`.
> 
> **Overview**
> Adds **`local-dylibs.sh`** so gitignored `local: true` cores (e.g. **virtualjaguar**, future **flycast-jitless**) can be **uploaded**, **fetched**, and **listed** via zipped release assets on **`Provenance-Emu/provenance-dylibs`** instead of living only on one machine.
> 
> The script **reads `cores.yml`** for `local: true` entries (awk, no PyYAML), supports a single **`filename`** dylib or the usual **`_ios` / `_tvos`** pair, zips dylibs like the buildbot layout, and uses **`gh`** with optional **`DYLIBS_REPO` / `DYLIBS_TAG`**. **`fetch`** fails the run if any expected asset is missing so builds don’t fail later on empty xcfilelists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ff5c9646781462c4dffb704bea9db4315b2b62f. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -154,6 +154,28 @@ jobs:
           fi
           echo "Submodule init complete."
 
+      # Cores marked `local: true` in cores.yml are custom Provenance builds that do
+      # not exist on the libretro buildbot, so get-modules.sh deliberately never
+      # downloads them. Nothing else fetched them either, and the failure mode is
+      # SILENT: make_frameworks_retroarch.sh iterates the dylibs that are present in
+      # modules/, so an absent one is simply never packaged — no error, no warning,
+      # just a core missing from the IPA. virtualjaguar (Jaguar) has been affected.
+      #
+      # Not hard-failing when DYLIBS_TOKEN is unset, because that would break every
+      # build the moment this lands. A named warning is still strictly better than
+      # today's silence: it says exactly which cores the artifact will lack.
+      # GITHUB_TOKEN cannot read another private repo, hence a dedicated PAT.
+      - name: Fetch locally-built core dylibs
+        env:
+          GH_TOKEN: ${{ secrets.DYLIBS_TOKEN }}
+        run: |
+          if [ -z "${GH_TOKEN}" ]; then
+            echo "::warning::DYLIBS_TOKEN not set — skipping locally-built cores. This IPA will be MISSING:"
+            ./CoresRetro/RetroArch/scripts/local-dylibs.sh list || true
+            exit 0
+          fi
+          ./CoresRetro/RetroArch/scripts/local-dylibs.sh fetch
+
       - name: Setup Xcode
         uses: maxim-lobanov/setup-xcode@v1.6.0
         with:
--- a/CoresRetro/RetroArch/scripts/local-dylibs.sh
+++ b/CoresRetro/RetroArch/scripts/local-dylibs.sh
@@ -1,0 +1,176 @@
+#!/usr/bin/env bash
+# local-dylibs.sh — publish/fetch the custom-built libretro dylibs.
+#
+# Some cores are NOT on the libretro buildbot: they are custom Provenance builds
+# (flycast-jitless, virtualjaguar). cores.yml marks them `local: true`, which stops
+# get-modules.sh downloading or pruning them — but says nothing about where they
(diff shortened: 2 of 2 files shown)


JoeMatt:
bugbot run

JoeMatt:
Extended this beyond a developer convenience — while wiring the token I found the gap is worse than the PR describes.

`build.yml` has **no provision for `local: true` cores at all**, and the failure mode is silent: `make_frameworks_retroarch.sh` iterates the dylibs *present* in `modules/`, so an absent one is never packaged — no error, no warning, just a core missing from the IPA. `virtualjaguar` is `local: true` on develop and appears in the generated xcfilelists, so **CI-built alphas have been shipping without the Jaguar core**.

Added a fetch step to `build.yml`. It deliberately does *not* hard-fail when `DYLIBS_TOKEN` is unset — that would break every build the moment this lands. Instead it warns and names exactly which cores the artifact will lack, which is still strictly better than the current silence. Once the secret exists it fetches for real.

To enable, a fine-grained PAT limited to the one private repo (least privilege — it only needs to read release assets):

1. https://github.com/settings/personal-access-tokens/new
2. Resource owner **Provenance-Emu**, repository access **Only select repositories** → `provenance-dylibs`
3. Repository permissions → **Contents: Read-only**
4. `gh secret set DYLIBS_TOKEN -R Provenance-Emu/Provenance`

Worth confirming empirically after the next alpha build that Jaguar is actually present — I inferred the omission from how the frameworks loop works rather than by inspecting a built IPA.

PR #3641, opened 2026-08-02
feat(cores): scaffold tvOS support for flycast-jitless, fix dylib name derivation

JoeMatt:
Wires the per-platform filename split so enabling tvOS later is a **one-line change**, without enabling it now against a dylib that can't serve tvOS.

## Why the iOS dylib can't stand in for tvOS

```
flycast-jitless_libretro.dylib   -> LC_BUILD_VERSION platform 2  (iOS)
2048_libretro_tvos.dylib         -> LC_BUILD_VERSION platform 3  (tvOS)
```

The buildbot publishes separate `ios-arm64` and `tvos-arm64` builds for **every** core, into separate `modules_compressed/iOS` and `/tvOS` dirs. A neutral `filename:` means the two platforms share a **name, not a file** — `get-modules.sh` extracts a different per-platform binary into that name. That distinction is invisible from `cores.yml`, which is what makes it look like one shared dylib.

## The scaffolding

`filename` is kept (iOS behaviour untouched) and `tvos_filename` added as an inert slot. Regenerating produces **byte-identical** lists while `tvos: false`.

Verified it actually works rather than merely being harmless — temporarily flipping `tvos: true`:

| | resolves to |
|---|---|
| iOS xcfilelist | `flycast-jitless_libretro.dylib` |
| tvOS xcfilelist | `flycast-jitless_libretro_tvos.dylib` |
| `local-dylibs.sh list` | correctly reports the tvOS asset MISSING |

So once you build a tvOS jitless dylib and `local-dylibs.sh upload` it, flipping `tvos: true` is the only remaining change.

## Two bugs found in local-dylibs.sh while doing it

**It only read `filename`.** A core using `ios_filename`/`tvos_filename` would have had its names mis-derived and been uploaded under a name the build never looks for. Now mirrors `CoreEntry.ios_filename()`/`tvos_filename()` precedence exactly.

**Tab-separated parsing was broken for empty fields.** Tab is IFS *whitespace*, so bash collapses runs of it into a single delimiter and drops empty fields — shifting every value left for any core omitting `filename`:

```
dylibs_for flycast-jitless flycast-jitless_libretro.dylib true false '' ''
                                      iosf=true  tvosf=false  ios=  tvos=
```

This silently produced *empty output* the moment the record grew past two fields. Separator moved to `|`. It also now honours the `ios`/`tvos` flags, so a platform-disabled core isn't reported missing — and doesn't fail `fetch` — for a dylib nothing will consume.

## Also

`flycast-jitless_libretro.dylib` (3.8 MB zipped) is now uploaded to the store, which it couldn't be before #3638 taught `cores.yml` about it.

---

> [!NOTE]
> **Low Risk**
> Build/CI scripting and YAML metadata only; no runtime app logic, with behavior unchanged for flycast-jitless while tvos remains false.
> 
> **Overview**
> Adds **inert** `tvos_filename` for **flycast-jitless** in `cores.yml` (with clearer docs on why `tvos` stays false until a tvOS Mach-O exists), so turning on tvOS later is mostly flipping `tvos: true` after uploading a tvOS dylib.
> 
> **`local-dylibs.sh`** is updated to match **`generate_core_lists.py`** naming: it reads `ios_filename` / `tvos_filename`, honors `ios` / `tvos` flags, and avoids duplicate tvOS entries when iOS and tvOS share the same neutral name. Parsing switches from tab to **`|`** so empty `filename` fields no longer break `IFS` and mis-assign columns.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9149531fb4d5b5e00f14963dc4d9abafed370be9. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CoresRetro/RetroArch/scripts/cores.yml
+++ b/CoresRetro/RetroArch/scripts/cores.yml
@@ -258,22 +258,16 @@ cores:
     # per-platform filename split (ios_filename / tvos_filename) since one neutral
     # name cannot carry both slices.
     ios: true
-    # tvos stays false until a tvOS-built jitless dylib exists. The iOS one cannot
-    # stand in: it is Mach-O platform 2 (iOS) and tvOS wants platform 3, which is
-    # also why the buildbot publishes separate ios-arm64 / tvos-arm64 builds for
-    # every core. A neutral `filename` means the two platforms share a NAME, not a
-    # file — get-modules.sh extracts a different per-platform binary into it from
-    # modules_compressed/<platform>/.
-    #
-    # tvos_filename below is scaffolding, inert while tvos is false. Once a tvOS
-    # build is uploaded via local-dylibs.sh, flipping tvos to true is the only
-    # change needed — the name is already wired through the generator and the
-    # store script.
-    tvos: false
+    # tvOS shares the SAME iOS-built dylib — no separate tvOS build exists or is
+    # needed. The Mach-O says platform 2 (iOS), but this loads fine on tvOS and is
+    # device-tested there via the App Store scheme in Xcode (JM, Aug 2026);
+    # Dreamcast is the premier tvOS core, so it must ship. If a future tvOS-specific
+    # build is ever wanted, add a tvos_filename override + upload that asset, and
+    # local-dylibs.sh / generate_core_lists.py already handle the split.
+    tvos: true
     appstore: true
     enabled: true
     filename: "flycast-jitless_libretro.dylib"
-    tvos_filename: "flycast-jitless_libretro_tvos.dylib"
     local: true
 
   - name: fmsx
--- a/CoresRetro/RetroArch/scripts/cores.yml
+++ b/CoresRetro/RetroArch/scripts/cores.yml
@@ -258,10 +258,22 @@ cores:
     # per-platform filename split (ios_filename / tvos_filename) since one neutral
     # name cannot carry both slices.
     ios: true
+    # tvos stays false until a tvOS-built jitless dylib exists. The iOS one cannot
+    # stand in: it is Mach-O platform 2 (iOS) and tvOS wants platform 3, which is
+    # also why the buildbot publishes separate ios-arm64 / tvos-arm64 builds for
+    # every core. A neutral `filename` means the two platforms share a NAME, not a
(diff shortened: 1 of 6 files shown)


JoeMatt:
Fixed in `09957bdc59`. The build break came from the autofix commit `03b3073889` ("Potential fix for pull request finding"), not the original change.

It inverted the awk parser's reset defaults:

```diff
- islocal = 0; ...; ios = "false"; tvos = "false"
+ islocal = 0; ...; ios = "true";  tvos = "true"
```

The parser only ever sets these **to** `"true"`, on a matching `ios: true` / `tvos: true` line — a `tvos: false` line never set them back. So with the defaults flipped, every core came out `ios=true tvos=true` regardless of what `cores.yml` actually says.

`flycast-jitless` is deliberately `tvos: false` (its dylib is iOS-only, Mach-O platform 2). `fetch` therefore demanded `flycast-jitless_libretro_tvos.dylib`, which doesn't exist in the store, and exited 1 — failing **both** archive jobs:

```
fetched flycast-jitless_libretro.dylib
WARNING: flycast-jitless_libretro_tvos.dylib.zip not found
fetched 3 dylib(s), 1 missing        → exit 1
```

Reverted, and hardened so the default can't matter again: explicit `ios: false` / `tvos: false` rules now set the flags too, so every value in `cores.yml` is honoured rather than inferred from an initial state. That makes the parser correct under either default.

Verified locally: `list` reports 3 dylibs (not 4), `fetch` exits 0.

JoeMatt:
Addressed. @copilot and @qodo were both right — `generate_core_lists.py:275-276` defaults `ios`/`tvos` to `True` and my parser defaulted them false, so a future `local: true` core omitting the flags would produce no dylib names: silently skipped by list/upload/fetch while the generator built it into both xcfilelists.

Defaults now match the generator.

Worth recording why this is safe when the identical change broke the build earlier (`03b3073889`): that attempt flipped the defaults while the parser could only ever set flags **to** `true`, so an explicit `tvos: false` never took effect. The explicit `ios: false` / `tvos: false` rules added in `09957bdc59` are what make the default irrelevant — every value in cores.yml is now read rather than inferred. The autofix had the right instinct and the wrong prerequisite.

PR #3642, opened 2026-08-02
feat(ci): build alphas with the sideload core set and sideload runtime rules

JoeMatt:
CI archives the **AppStore scheme** for alpha IPAs, so sideload users were getting App Store restrictions on two independent axes. This fixes both, gated so real ASC/TestFlight builds are untouched.

## 1. Embedded cores

The AppStore target's two RetroArch phases hardcode the restricted set (`get-modules.sh -appstore` + `urls-appstore*.txt` filter). After #3638 stopped `cores.yml` lying about flycast, CI alphas **lost JIT flycast entirely** — correct for the App Store, wrong for the audience actually receiving these IPAs, who can run JIT cores.

Both phases now branch on `PV_ALPHA_SIDELOAD=YES`, passed only by `build.yml` for alpha builds:

| Build | Module set |
|---|---|
| CI alpha (`IS_ALPHA=true`) | full sideload (`urls*.txt`) — JIT flycast, dolphin, … |
| Local Xcode archive / release.sh / testflight.yml | appstore set, unchanged |

Sandboxing checked: `ENABLE_USER_SCRIPT_SANDBOXING = NO` on the app target (the 27 `YES` blocks are extensions without these phases), so the sideload run writing more dylibs than the declared appstore output xcfilelist is harmless — the stale declaration only affects incremental builds, which clean CI runs never exercise.

## 2. Runtime rules

`AppState.isAppStore` keys off `PVAppType` in Info.plist, and the archived target's plist says `standard.appstore` — so sideloaded alphas enforced **every** runtime App Store gate: JIT manager, `appStoreDisabled` core filtering, upsells. The alpha patch step now sets `PVAppType=standard` in the plists the scheme *actually* reads (`*-AppStore-Info.plist`).

Found while doing it: the old patch step targeted `Provenance/Info.plist` + `ProvenanceTV/Info.plist` — the first isn't the archived target's `INFOPLIST_FILE` and the second **doesn't exist** — so the "Provenance Alpha" display-name patch has been a silent no-op since it shipped. It now genuinely applies.

## Verification

Dry-run validated locally: pbxproj parses, YAML parses, both shell branches resolve to the right filter per platform, PlistBuddy commands verified against copies of the real plists.

OWNER PRs don't trigger the archive matrix, so end-to-end proof lands with the **first develop alpha after merge**: the IPA should contain `dolphin_libretro.dylib` / `flycast_libretro.dylib`, and the installed app should report `isAppStore=false`, `distributionChannel=sideload`.

---

> [!NOTE]
> **Medium Risk**
> Changes affect which cores ship in develop alphas and runtime App Store gating via plist patching; mis-gating on non-alpha builds would be serious, but the flag is CI-only for develop pushes.
> 
> **Overview**
> **Develop alpha IPAs** are built with the App Store scheme but distributed to sideload users. This PR aligns **what gets embedded** and **how the app behaves at runtime** with that audience, without changing TestFlight/App Store archives.
> 
> **CI (`build.yml`)** now passes `PV_ALPHA_SIDELOAD=YES` only when `IS_ALPHA=true`, alongside the existing bundle suffix. The alpha plist step targets `Provenance-AppStore-Info.plist` and `ProvenanceTV-AppStore-Info.plist` (the plists the archived target actually uses) instead of the wrong/missing paths—so **“Provenance Alpha”** display name and **`PVAppType=standard`** actually land in the IPA. That flips sideload alphas off `AppState.isAppStore` behavior (JIT manager, App Store core filtering, upsells).
> 
> **Xcode build scripts** in `project.pbxproj` branch on `PV_ALPHA_SIDELOAD`: **Get RetroArch Modules** and **Generate Frameworks** use the full sideload URL lists and `get-modules.sh` without `-appstore`, matching each other so dylibs and packaged frameworks stay in sync. Unset flag keeps the existing App Store–restricted set for local/release builds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0ab9a5daa98d8eb7ead1dc92fa2d896219365a20. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


JoeMatt:
Both findings were real, fixed.

**@copilot / @qodo — `appletvsimulator`:** not hypothetical. `get-modules.sh:87` and `make_frameworks_retroarch.sh:80` already match `appletvos|appletvsimulator`, and get-modules.sh carries this comment:

> `# ... Only checking appletvos wrongly downloads iOS dylibs`

I duplicated the exact mistake it warns about. With mismatched filters `make_frameworks` rejects every `*_tvos.dylib` and exits non-zero with "0 frameworks created". Both branches now use one `case` matching the shell scripts. Verified all 8 platform × sideload combinations resolve correctly.

**@sentry — PlistBuddy `Set` on a missing key:** right, and my `|| echo ::warning::` made it worse than you flagged — the build would continue and ship an alpha still reporting `standard.appstore`, re-enabling every runtime App Store gate for sideload users. That's the entire point of this PR, so it shouldn't degrade to a warning. Now Delete-then-Add (tolerates missing, always creates) with `::error::` + `exit 1` on failure. Verified against a plist lacking the key entirely.

**@qodo — maintainer manual-apply note:** declining, same as #3633. That rule exists because the `github-actions` bot lacks the `workflows` permission; this PR is authored under maintainer credentials, so the note would tell a reader to wait for a step that isn't required.

Issue #3612, opened 2026-03-30
Logs export share


JoeMatt:
`RetroLogView` ties in with our `PVLogging` system.

I would like to add the ability to export and share logs in a common format that Provenance can also openURL/import to view

it proably should be a zip bundle of some sort with crash logs, text logs, retroarch logs etc

it should provide options for what to include, and either as a zipped bundle or a single text file for just logs with some header timestamp data or frontmatter or something i dunno

you should be able to export using the native ios share apis

you should be abale to import a log and view it or open a log file by extension or mime type from the app bootup and open the log viewer with those logs and an indicator about viewing another log session.

Perhaps we can also add new log menu items to show retroarch logs, which there could be single or timestamped logs in RetroArch/logs

also quick access to app and retroarch log settings if possible

also, if pvlogging is logging to date stamped files that roll over / rotate, can we view them in the log viewer too and also delete or share them?

JoeMatt:
@claude spike this out and make work PRs

JoeMatt:
## Status — 5 of 6 asks shipped

| Ask | Status |
|---|---|
| Export & share logs (native share sheet) | ✅ #3613 (merged Apr 2) |
| Zip bundle **or** single text file, with header/frontmatter | ✅ #3613 — `device_info.txt` header, `.txt` or `.zip` |
| Options for what to include | ✅ #3613 — app logs / device info / RetroArch logs |
| RetroArch logs menu (`RetroArch/logs`) | ✅ #3614 — Settings › Debug › RetroArch Logs |
| Rotating date-stamped PVLogging files — view / share / delete | ✅ #3615 — `PVLogFileManager` + session browser |
| **Import a log and view it, with an indicator** | ✅ **#3645** (just opened) |
| Open a log by file extension / MIME at app bootup | ❌ still open — see below |

### #3645 — import & view

Import button in the `RetroLogView` header takes `.txt`, `.log`, or an exported `.zip` bundle. The imported session replaces the live list, with a banner naming the file and a **LIVE LOGS** button to go back. Levels are colour-parsed best-effort; search filters imported lines too. iOS/Catalyst only — tvOS has no document picker.

### The one remaining piece

[...]

PR #3645, opened 2026-08-05
[Agent] feat(logs): import & view exported log files (#3612)

JoeMatt:
Completes the import half of #3612. Export (#3613), the RetroArch log browser (#3614), and `PVLogFileManager` (#3615) already landed; importing a log back in was the remaining gap.

## What it does

- **Import button** (`square.and.arrow.down`) in the `RetroLogView` header, next to Export
- Accepts **`.txt`, `.log`, and exported `.zip` bundles** — ZIPs are unpacked via ZIPFoundation (already a PVUIBase dependency), with `device_info.txt` ordered first so the header reads at the top
- Imported session **replaces the live log list**, with a banner naming the file and a **LIVE LOGS** button to return
- Best-effort `[LEVEL]` parsing so imported lines keep their colour coding
- Search still filters imported lines; Clear is disabled while viewing an import
- Failure surfaces a plain alert rather than failing silently

## Platform

iOS / Mac Catalyst only — tvOS has no document picker, so the button, `.fileImporter`, and `UniformTypeIdentifiers` import are all behind `#if !os(tvOS)`. The banner and row rendering compile on both.

## Validation

- `swiftc -parse` clean on both changed files
- `swiftlint` — no errors. New subviews were moved into an `extension RetroLogView` so `type_body_length` went **down** (638 → 587 lines, under the 600 error threshold); the remaining warning is pre-existing debt on this file.

## Not included

Opening a log file from outside the app (Files.app / share sheet → Provenance) is **still open**. It needs `CFBundleDocumentTypes` across 8 Info.plists plus routing in `handle(fileURL:)`. Worth a deliberate decision first: claiming `public.plain-text` / `public.zip-archive` would make Provenance volunteer to open *every* text file and ZIP from other apps, and risks intercepting ROM ZIP imports. A custom exported UTI is the safer path.

---

> [!NOTE]
> **Low Risk**
> Scoped to the in-app log viewer and user-selected files with security-scoped access; no auth, persistence, or global file-type registration changes.
> 
> **Overview**
> Adds **log file import** to `RetroLogView` on iOS and Mac Catalyst (not tvOS): a header **import** control opens a `.fileImporter` for `.txt`, `.log`, and exported `.zip` bundles.
> 
> `RetroLogViewModel` gains import support via ZIPFoundation: plain UTF-8 files or ZIPs whose `.txt`/`.log` entries are merged (`device_info.txt` first), with `[LEVEL]` tags parsed for line coloring. An imported session **replaces** the live list behind a banner with **LIVE LOGS** to dismiss; search applies to imported lines and **Clear** is disabled while viewing an import. Failures show an **Import Failed** alert.
> 
> Imported-session banner and line rows live in a `RetroLogView` extension; picker, importer, and UTTypes stay behind `#if !os(tvOS)`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 25a65442611e69411798a17727d97a1529eb4634. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changelog/3645.md
+++ b/.changelog/3645.md
@@ -1,0 +1,2 @@
+### Added
+- **Log Import** — the retrowave log viewer can now import a previously exported `.txt`/`.log` file or `.zip` bundle and display it in place of live logs, with a banner to return to live logs.
--- a/.github/workflows/agent-validation.yml
+++ b/.github/workflows/agent-validation.yml
@@ -186,13 +186,18 @@ jobs:
       - name: Select Xcode
         run: sudo xcode-select -s /Applications/Xcode_26.3.app
 
+      # `set -o pipefail` is required: the default shell here is `bash -e`
+      # (no pipefail), so piping into `tail` masked the real exit status and
+      # these steps reported success even when the build or tests failed.
       - name: Build module
         run: |
+          set -o pipefail
           cd ${{ matrix.module }}
           swift build 2>&1 | tail -20
 
       - name: Test module
         run: |
+          set -o pipefail
           cd ${{ matrix.module }}
           swift test 2>&1 | tail -40
 
--- a/PVLogging/Sources/PVLogging/LogFileParsing.swift
+++ b/PVLogging/Sources/PVLogging/LogFileParsing.swift
@@ -1,0 +1,62 @@
+//
+//  LogFileParsing.swift
+//  PVLogging
+//
+//  Parsing helpers for reading previously-exported log files back in.
+//  Lives here rather than in the UI layer so it is testable without a
+//  full workspace build — PVLogging builds and tests standalone.
+//
+
+import Foundation
+
+/// Parses plain-text log files produced by Provenance's log exporter (or any
+/// log whose lines carry a `[LEVEL]` tag) back into structured lines.
+public enum LogFileParsing {
+
+    /// A single parsed line of a log file.
+    public struct ParsedLine: Identifiable, Equatable, Sendable {
+        /// Zero-based index of the line within the file. Stable and unique,
+        /// which makes it usable directly as a SwiftUI `ForEach` identity.
+        public let id: Int
+        /// The raw line text, unmodified.
+        public let text: String
+        /// Level parsed from a `[LEVEL]` tag, or `nil` when the line carries none.
+        public let level: LogLevel?
+
+        public init(id: Int, text: String, level: LogLevel?) {
+            self.id = id
+            self.text = text
+            self.level = level
+        }
+    }
+
+    /// Splits `text` into parsed lines, preserving order and blank lines.
+    ///
+    /// CRLF is normalised first: `.newlines` is a character set, so it would
+    /// otherwise treat `\r\n` as two breaks and insert a spurious blank line
+    /// between every line of a Windows-generated log.
+    /// - Parameter text: Full contents of a log file.
+    /// - Returns: One `ParsedLine` per line, indexed from zero.
+    public static func parseLines(_ text: String) -> [ParsedLine] {
+        text
+            .replacingOccurrences(of: "\r\n", with: "\n")
+            .components(separatedBy: .newlines)
+            .enumerated()
+            .map { ParsedLine(id: $0.offset, text: $0.element, level: level(in: $0.element)) }
+    }
+
+    /// Best-effort parse of a `[LEVEL]` tag from a single log line.
+    ///
+    /// Matches the uppercase tag emitted by the exporter (`[ERROR]`, `[WARNING]`,
+    /// `[INFO]`, `[DEBUG]`, `[VERBOSE]`). Returns `nil` for untagged lines so
+    /// callers can render them neutrally rather than guessing.
+    public static func level(in line: String) -> LogLevel? {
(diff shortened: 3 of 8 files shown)


JoeMatt:
Addressed all review feedback in `afeafed18b`.

| Finding | Reviewer | Fix |
|---|---|---|
| `.textSelection(.enabled)` breaks tvOS build | Copilot | **Real build break** — now behind `#if !os(tvOS)`. Verified with `swiftc -parse -target arm64-apple-tvos17.0`. |
| Copy/Export use live logs while viewing an import | cursor | Copy now copies what's on screen; Export is disabled during an import (it exports the live session, so sharing it would be misleading). |
| Auto-scroll targets IDs not in the imported list | cursor, Copilot | `handleAutoScroll` early-returns while an import is open — imported rows use `Int` IDs, live rows use `UUID`. |
| `importableExtensions` is dead API | cursor, Copilot, qodo | Removed. |
| `importLog(from:)` publishes off-main | Copilot | Marked `@MainActor`. |
| Import button has no accessibility label | Copilot | Added label + hint. |
| No tests for import/ZIP/level parsing | Copilot, qodo | Added `RetroLogImportTests` (12 cases): plain-text and ZIP import, `[LEVEL]` parsing, `device_info.txt` ordering, non-text archive entries ignored, search filtering, unique line IDs, and that a **failed import doesn't clobber the live session**. |

Validation: `swiftc -parse` clean for **both** iOS and tvOS targets, `swiftlint` no errors. I could not execute the tests locally — PVUIBase is Tier 6 and needs a full workspace build.

JoeMatt:
### Re: the repeated cursor/Copilot comments

The bots are re-posting their original review on each push rather than diffing current state. All of the repeated findings were fixed in `afeafed18b` and survive the rebase — verified against `git show HEAD`, not from memory:

| Repeated finding | Actual state at `c6a2a2de9e` |
|---|---|
| "Copy exports live not imported" | `copyFilteredLogsToClipboard()` branches on `importedSession != nil` and copies `displayedImportedLines` |
| "Auto-scroll ignores imported session" | `handleAutoScroll` opens with `guard viewModel.importedSession == nil else { return }` |
| "Unused `importableExtensions`" | Constant deleted — `git grep importableExtensions` returns nothing (that comment is also still anchored to the original commit `25a6544261`) |
| Export can share the wrong logs | Export button is `.disabled(viewModel.importedSession != nil)` |
| `.textSelection` breaks tvOS | Behind `#if !os(tvOS)`; verified with `swiftc -parse -target arm64-apple-tvos17.0` |

### New in `c6a2a2de9e` — parsing moved to PVLogging

`PVUIBaseTests` never actually run in CI: PVUI's package graph is unresolvable (stale `PrefirePlugin` product name — Prefire 2.10.0 ships `PrefireTestsPlugin`/`PrefirePlaybookPlugin` — plus ~12 macOS platform mismatches across sibling manifests). So the pure parsing moved to `PVLogging.LogFileParsing`, which is Tier 1, builds standalone, and is already in the SPM matrix.

`cd PVLogging && swift test` → **67 tests in 9 suites passed.**

Those tests immediately caught a real bug: `.newlines` is a character set, so CRLF split into two breaks and a Windows-generated log imported double-spaced. `parseLines` now normalises CRLF first.

JoeMatt:
### Review threads cleared — 13 unresolved → 0

Three of them were **real findings I had not addressed**, fixed in `a890983de0`:

1. **Import ran on the main thread** (qodo). The `.fileImporter` callback called `importLog(from:)` synchronously — file I/O, ZIP extraction and full parsing before returning — and my own `@MainActor` annotation *guaranteed* it ran on main. Parsing now runs in a detached task; only the final `importedSession` publish hops back. Fair catch.
2. **Unbounded reads** (qodo). No size limit anywhere: a large log or a highly-compressed archive expanded `Data` → `String` → `[ImportedLogLine]` unchecked. Added a 64 MB cap, enforced against the file size up front and against each ZIP entry's declared `uncompressedSize` *before* extraction, so a compression bomb is refused rather than expanded. Covered by a new test.
3. **`UIPasteboard` on non-UIKit platforms** (Copilot). Pre-existing, but I'd added a second usage. The guard was `#if !os(tvOS)`, which doesn't exclude native macOS/watchOS — both declared by PVUIBase. Now `#if !os(tvOS) && canImport(UIKit)`.

While fixing (2) I caught a compile error the bots didn't: ZIPFoundation's `uncompressedSize` is `UInt64`, so comparing it against an `Int` budget wouldn't build. `swiftc -parse` can't catch that (syntax only), so it's worth flagging that PVUIBase still has **no type-checking in CI** — see below.

The remaining threads were duplicate re-posts of findings already fixed in `afeafed18b`; I verified each against `git show HEAD` before resolving rather than taking the bots' word or my own changelog.

### Related: CI couldn't fail

`d1115aa775` — the SPM matrix steps run `swift build|test 2>&1 | tail -N` under the default `bash -e` (**no** `pipefail`), so they returned `tail`'s exit status and reported success even when tests failed. Added `set -o pipefail`. Heads-up that this may turn previously-green PRs red for the other matrix modules; that's the gate working, but it's a real change in signal.

PR #3646, opened 2026-08-05
fix(icloud): simulator entitlement check crashed unentitled builds at launch

JoeMatt:
## The bug

`iCloudConstants.isCloudKitEntitlementPresent` decided the simulator case from Info.plist:

```swift
if let containers = Bundle.main.infoDictionary?["NSUbiquitousContainers"] ... { return true }
```

`NSUbiquitousContainers` is a **declaration**, not an entitlement, and it's always present in `Provenance-Info.plist`. So the check returned `true` unconditionally on simulator — including for builds carrying **no entitlements at all**.

`iCloudConstants.container` then ran `CKContainer(identifier:)`, which **traps** (it can't be caught) when the container isn't entitled. The app dies at launch, before any UI:

```
CKContainer.__allocating_init(identifier:)
static iCloudConstants.container            iCloudConstants.swift:96
one-time initialization for shared          CloudSyncManager.swift:77
DirectoryWatcher.handleFileSystemEvent      DirectoryWatcher.swift:895
```

Any file-system event triggers it, so it fires during first-launch import scanning.

**Who it hits:** simulator builds signed without entitlements — CI (`CODE_SIGNING_ALLOWED=NO`), ad-hoc re-signs, screenshot automation. Developers building from Xcode with a team have real entitlements, so it never reproduced for them.

## The fix

Read the entitlements the linker embeds in `__TEXT,__entitlements` for simulator builds, via `_dyld_get_image_header` + `getsectiondata`, and look for `com.apple.developer.icloud-container-identifiers`. No section → no entitlements → `false`.

Two proxies were tried first and **both were wrong** — thanks to review for catching them:

| Attempt | Why it failed |
|---|---|
| `SecTaskCopyValueForEntitlement` (mirroring the macOS branch) | `SecTask.h` ships in the **macOS SDK only** — it doesn't exist for simulator targets, so this could not compile. `import Security` in this file is also gated to macOS/Catalyst. |
| `FileManager.ubiquityIdentityToken` | Reflects the **ubiquity** entitlement + account state, not the CloudKit container entitlement that gates `CKContainer`. A build holding one but not the other would still trap. |

Simulator apps aren't code-signed the way device builds are, which is why both the macOS `SecTask*` path and the device `embedded.mobileprovision` path miss these entitlements — the Mach-O section is where they actually live.

## Also gated

`SyncProviderFactory` (tvOS and non-tvOS branches) and `CloudSyncStatusView.checkCloudKitAvailability` constructed `CKContainer` directly and would trap the same way. They now go through `iCloudConstants.container`, falling back to the iCloud Documents syncer / skipping the check when unentitled.

**Not** included: `CloudKitDiagnosticView:477` and `CloudKitRecordsViewModel:130` hold non-optional stored properties feeding a derived `CKDatabase` (~7 and ~15 use sites). Making those optional cascades through view state — a wider refactor than belongs in a crash fix, especially since PVUIBase has no type-checking in CI (#3647). Tracked in **#3648**. Those are diagnostic/settings screens, so they can't cause the launch crash this PR fixes, but they would still trap if opened in an unentitled build.

## Verification

- Mach-O logic **type-checked standalone against the `iphonesimulator` SDK** — `-parse` is syntax-only and would not have caught the missing `SecTask` symbol, which is exactly how that slipped through.
- `swiftc -parse` + `swiftlint` clean on all changed files (`CloudSyncStatusView` parse-checked against the tvOS target, since its change is inside `#if os(tvOS)`).
- Full `Provenance (AppStore)` iOS Simulator build for a real type-check of the whole change.

## Risk

The behaviour change is confined to `#if targetEnvironment(simulator)`; device, Catalyst and App Store paths are untouched. Every `iCloudConstants.container` call site already uses `if let`/`guard let` (verified — no force-unwraps in PVLibrary or PVUI), so an unentitled build degrades to sync-disabled instead of crashing. An entitled simulator build still reports `true` and gets CloudKit as before.

### DIFF
--- a/PVLibrary/Sources/PVLibrary/Cloud Sync/SyncProviderFactory.swift
+++ b/PVLibrary/Sources/PVLibrary/Cloud Sync/SyncProviderFactory.swift
@@ -33,16 +33,21 @@ public class SyncProviderFactory {
         DLOG("iCloudSync=\(iCloudSyncEnabled), iCloudSyncMode=\(syncMode.description)")
         
 #if os(tvOS)
-        // Get the container identifier from the bundle
-        let containerIdentifier = iCloudConstants.containerIdentifier
-        let container = CKContainer(identifier: containerIdentifier)
+        // Use the entitlement-gated container: `CKContainer(identifier:)` traps when the
+        // container isn't entitled, so constructing it directly crashes unentitled builds.
+        guard let container = iCloudConstants.container else {
+            WLOG("CloudKit container unavailable (no entitlement) — falling back to iCloud Documents syncer")
+            return iCloudContainerSyncer(directories: directories, notificationCenter: notificationCenter, errorHandler: errorHandler)
+        }
         return CloudKitSyncer(container: container, directories: directories, notificationCenter: notificationCenter, errorHandler: errorHandler)
 #else
         // Return the appropriate syncer based on the mode
         if syncMode.isCloudKit {
-            // Get the container identifier from the bundle
-            let containerIdentifier = iCloudConstants.containerIdentifier
-            let container = CKContainer(identifier: containerIdentifier)
+            // Entitlement-gated: see the tvOS branch above.
+            guard let container = iCloudConstants.container else {
+                WLOG("CloudKit container unavailable (no entitlement) — falling back to iCloud Documents syncer")
+                return iCloudContainerSyncer(directories: directories, notificationCenter: notificationCenter, errorHandler: errorHandler)
+            }
             DLOG("Creating CloudKit syncer based on iCloudSyncMode=\(syncMode.description)")
             return CloudKitSyncer(container: container, directories: directories, notificationCenter: notificationCenter, errorHandler: errorHandler)
         } else {
--- a/PVLibrary/Sources/PVLibrary/Cloud Sync/iCloud/iCloudConstants.swift
+++ b/PVLibrary/Sources/PVLibrary/Cloud Sync/iCloud/iCloudConstants.swift
@@ -41,21 +41,28 @@ public enum iCloudConstants {
     /// - iOS / tvOS device: parses `embedded.mobileprovision` when present (sideloaded /
     ///   AdHoc / TestFlight). App Store builds strip the provisioning profile, so absence
     ///   of the file is treated as "entitlement present" (App Store validation guarantees it).
-    /// - Simulator: reads the live code signature via `SecTaskCopyValueForEntitlement`.
-    ///   Simulator builds carry no `embedded.mobileprovision`, and may carry no
-    ///   entitlements at all, so nothing may be assumed here.
+    /// - Simulator: uses `FileManager.ubiquityIdentityToken`, which is nil without the
+    ///   ubiquity entitlement. Simulator builds carry no `embedded.mobileprovision` *and*
+    ///   may carry no entitlements at all, so nothing may be assumed here — and the
+    ///   macOS `SecTask*` path is unavailable (that header ships only in the macOS SDK).
     public static let isCloudKitEntitlementPresent: Bool = {
         #if targetEnvironment(simulator)
         // `NSUbiquitousContainers` is an Info.plist *declaration*, not an entitlement.
         // It is always present in this bundle, so keying off it returned true even for
         // simulator builds signed without entitlements (CI's CODE_SIGNING_ALLOWED=NO,
         // ad-hoc re-signs, screenshot automation). `CKContainer(identifier:)` then traps
-        // on first access and the app dies at launch. Read the real code signature instead
-        // — the same source CloudKit itself validates against.
-        guard let task = SecTaskCreateFromSelf(nil),
-              let value = SecTaskCopyValueForEntitlement(task, "com.apple.developer.icloud-container-identifiers" as CFString, nil),
-              let containers = value as? [String] else { return false }
-        return !containers.isEmpty
+        // on first access and the app dies at launch.
+        //
+        // The code signature can't be read here the way the macOS branch does it:
+        // `SecTask.h` ships in the macOS SDK only, so `SecTaskCreateFromSelf` /
+        // `SecTaskCopyValueForEntitlement` don't exist for iOS/tvOS simulator targets.
+        //
+        // `ubiquityIdentityToken` is nil unless the process actually holds the ubiquity
+        // entitlement, is a plain Foundation API available on every platform, and — unlike
+        // `CKContainer(identifier:)` — returns rather than trapping. An entitled simulator
+        // build with a signed-in account still gets CloudKit; an unentitled one degrades
+        // to sync-disabled instead of crashing at launch.
+        return FileManager.default.ubiquityIdentityToken != nil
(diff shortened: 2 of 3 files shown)


JoeMatt:
The review was right on all three counts — the first version **could not have compiled**. Fixed in `1a7260cf62`.

### 1 & 2. `SecTask*` doesn't exist on iOS (Sentry HIGH, Copilot, qodo)

Verified against the SDKs rather than taking it on faith:

```
$ ls $(xcrun --sdk iphonesimulator --show-sdk-path)/.../Security.framework/Headers/SecTask.h
No such file or directory

$ ls $(xcrun --sdk macosx --show-sdk-path)/.../Security.framework/Headers/SecTask.h
.../MacOSX26.5.sdk/System/Library/Frameworks/Security.framework/Headers/SecTask.h
```

`SecTask.h` ships in the macOS SDK only. And Copilot's separate point stands too — `import Security` in this file is gated to `#if os(macOS) || targetEnvironment(macCatalyst)`, so the simulator branch had no import either way. Two independent reasons it would have failed to build.

**Replaced with `FileManager.ubiquityIdentityToken`:** plain Foundation, available on every platform, nil without the ubiquity entitlement, and it *returns* instead of trapping — which is the whole point, since `CKContainer(identifier:)` traps and can't be caught.

Trade-off worth stating: an entitled simulator build that is **signed out of iCloud** now reports `false` and runs with sync disabled. That's a behaviour change from the old always-`true` path, but it fails safe rather than crashing, and sync couldn't work without an account anyway.

### 3. Call sites bypassing the guard (qodo)

Correct — five of them. `SyncProviderFactory`'s two are now gated behind `iCloudConstants.container` and fall back to the iCloud Documents syncer when unentitled (both the tvOS and non-tvOS branches).

The remaining three are **not** fixed here, deliberately:

- `CloudKitDiagnosticView.swift:477` — `private let container` stored property feeding `privateDatabase`
- `CloudKitRecordsViewModel.swift:130`
- `CloudSyncStatusView.swift:209`

Making those optional cascades through view state and every downstream use, which is more churn than belongs in a crash fix during a release window. They're all in CloudKit diagnostic/settings UI rather than the launch path, so they can't cause the launch crash this PR is about — but in an unentitled build they would still trap if a user opens those screens. Worth a follow-up.

### Verification

`swiftc -parse` clean on both changed files; a full `Provenance (AppStore)` iOS Simulator build is running now for a real type-check, since `-parse` is syntax-only and would not have caught the missing-symbol problem that started this.

JoeMatt:
Both new findings were correct. Fixed in `eeb478b30c`.

### `ubiquityIdentityToken` was the wrong signal

Right — it reflects the **ubiquity** entitlement and account state, not `com.apple.developer.icloud-container-identifiers`, which is what actually gates `CKContainer(identifier:)`. A build holding one but not the other would still trap. That's the same class of mistake as the original `NSUbiquitousContainers` check this PR set out to fix: reaching for a proxy instead of the thing itself. Two proxies in a row now, so:

**Now reads the real entitlements blob.** Simulator apps aren't code-signed like device builds — the linker embeds entitlements in `__TEXT,__entitlements`. `_dyld_get_image_header(0)` + `getsectiondata` pulls that section and looks for the container list; no section (a build with no entitlements at all) reports `false`. That's the same data CloudKit validates against, so it can't false-positive the way either proxy could.

Verified by **type-checking the logic standalone against the `iphonesimulator` SDK**:

```
xcrun swiftc -typecheck -sdk $(xcrun --sdk iphonesimulator --show-sdk-path) \
  -target arm64-apple-ios17.0-simulator machotest.swift   → exit 0
```

`-parse` alone is syntax-only, which is precisely why the `SecTaskCreateFromSelf` call sailed through my checks and had to be caught in review. Type-checking against the target SDK is the check that would have caught it, so I've adopted it here.

### Stale PR description

Also correct, and fixed — the description still described the `SecTaskCopyValueForEntitlement` approach that no longer exists in the diff. Rewritten to match the code, including a table of both failed attempts and why each was wrong.

### Remaining `CKContainer` bypasses

`SyncProviderFactory` (both branches) and `CloudSyncStatusView.checkCloudKitAvailability` are now gated. `CloudKitDiagnosticView:477` and `CloudKitRecordsViewModel:130` are **not**, deliberately — they hold non-optional stored properties feeding a derived `CKDatabase` (~7 and ~15 use sites), so making them optional cascades through view state. That's a wider refactor than belongs in a crash fix, especially with no type-checking available for PVUIBase (#3647). Tracked in **#3648** with a suggested shape (an `isCloudKitUnavailable` flag driving an explanatory empty state, rather than 22 scattered guards).

I've resolved the three `SecTask` threads since that code is gone.

JoeMatt:
Latest round addressed in `4a89ad1` — including one finding that turned out to be the most important on this PR.

### CRITICAL (Sentry) — pre-existing, on the **device** path

`_cloudKitEntitlementFromProvisioningProfile()` did:

```swift
String(raw[plistStart.lowerBound ..< plistEnd.upperBound]) + "</plist>"
```

`plistEnd.upperBound` is already past the closing tag, so this produced `</plist></plist>`. Malformed XML → parse fails → function returns `nil` → the caller's `?? true` reads that as "App Store build, assume entitled."

**The device entitlement check has therefore never worked.** Sideloaded builds without the CloudKit entitlement have been falling through to `true` and trapping in `CKContainer(identifier:)` — exactly the crash this PR fixes for simulator, sitting on the device path the whole time. Not introduced here; found because this PR drew attention to the function.

### Two regressions I introduced — both correct catches

- **Stuck status.** My `CloudSyncStatusView` guard returned without touching UI state, leaving the view on its initial "Checking sync status…" forever while the header said unavailable. Now sets `isAvailable = false` and a terminal message.
- **Invalid tvOS fallback.** I fell back to `iCloudContainerSyncer`, but `iCloudSyncMode.iCloudDrive` is `#if !os(tvOS)` and its own subtitle says "Not supported on tvOS" — I handed back a provider the platform declares invalid. Now returns `nil`. `createSyncProvider` has **zero callers** repo-wide, so the `SyncProvider?` signature change is contained.

### The remaining thread — view models still bypassing the guard

Correct, and still deferred rather than dismissed. `CloudKitDiagnosticView:477` and `CloudKitRecordsViewModel:130` hold non-optional stored properties feeding a derived `CKDatabase` (~7 and ~15 use sites), so making them optional cascades through view state. Tracked in **#3648** with a suggested shape (an `isCloudKitUnavailable` flag driving an explanatory empty state, rather than 22 scattered guards).

To be straight about why it isn't in this PR: PVUIBase has no type-checking in CI (#3647), so a 20+ site refactor there can only be validated by a full simulator archive build — and I'd rather land the launch-crash fix first than bundle an unverifiable refactor into it. Those are diagnostic screens, unreachable before launch, so they can't cause the crash this PR targets.

### Verification

`swiftc -parse` (iOS + tvOS targets) and `swiftlint` clean on all three files. The Mach-O entitlement read was type-checked standalone against both the `iphonesimulator` and `appletvsimulator` SDKs. A full workspace build succeeded, though I'll re-run it on a stable tree — my own edits landed mid-build, so that result isn't as clean a signal as I'd like.

Issue #3648, opened 2026-08-05
CloudKit diagnostic view models trap on unentitled builds (CKContainer bypass)


JoeMatt:
## Summary

Two CloudKit settings view models construct `CKContainer(identifier:)` directly, bypassing the entitlement-gated `iCloudConstants.container`. `CKContainer(identifier:)` **traps** (not throws) when the container isn't in the process's entitlements, so opening these screens in a build without the CloudKit entitlement crashes the app.

This mainly affects **sideloaded builds** signed with a different team, where the entitlement is absent but the UI is still reachable from Settings.

## Sites

| File | Line | Shape |
|---|---|---|
| `PVUI/Sources/PVUIBase/SwiftUI/CloudKitDiagnosticView.swift` | 477 | `private let container = CKContainer(identifier:)` → `privateDatabase` in `init()`; ~7 use sites |
| `PVUI/Sources/PVSwiftUI/Settings/ViewModels/CloudKitRecordsViewModel.swift` | 130 | `self.container = CKContainer(identifier:)` in `init()` → `database`; ~15 use sites |

## Already fixed elsewhere

- `SyncProviderFactory` (both branches) and `CloudSyncStatusView.checkCloudKitAvailability` — gated in #3646. Those were function-local or returned a value, so the guard was contained.

## Why these two were deferred

Both hold **non-optional stored properties** feeding a derived `CKDatabase`. Making them optional cascades through every use site and the surrounding `@Published` view state. That's a wider change than belonged in a crash fix during a release window — and, per #3647, **PVUIBase and PVSwiftUI have no type-checking in CI**, so a 20-site refactor there can't be verified short of a full simulator archive build.

## Suggested approach

Make `container`/`database` optional and add an `isCloudKitUnavailable` flag the views can render as an explanatory empty state, rather than scattering `guard` at 22 call sites. That also gives sideloaded users a clear message instead of a dead screen.

Blocked-ish on #3647 for a sane verification loop.

PR #3649, opened 2026-08-05
fix(release): stop editing Build.xcconfig so it can't be left dirty

JoeMatt:
## Symptom

Found the working tree in this state after a killed release run:

```diff
-CURRENT_PROJECT_VERSION = 9951
+CURRENT_PROJECT_VERSION = 1785899074
```

plus an orphaned `Build.xcconfig.release-bak` (byte-identical to the committed file). `1785899074` is a Unix timestamp, not the commit count the file's own comment calls for — the epoch fallback had been written into a **tracked** file and left there.

## Why the existing trap didn't save it

`release.sh` edited the xcconfig and undid it from an `EXIT` trap. That's sound until the process can't run a trap:

- `SIGKILL` — nothing runs, by definition
- a wedged `xcodebuild` swallowing `SIGINT` — the script's own comment already documented this ("mashing Ctrl-C will do nothing")

Then the dirty file persists until someone happens to run the script again and the self-heal fires. In the meantime any local build stamps a timestamp build number, and the change can ride along in an unrelated commit.

## Fix

Don't write the file. The edit was **redundant**:

- All three real `xcodebuild` invocations — `archive` and both `-exportArchive` — already pass `CURRENT_PROJECT_VERSION="$BUILD_NUMBER"` on the command line, which outranks the xcconfig
- `Info.plist` resolves `CFBundleVersion` from the **build setting**, not the file — see the comment in `Scripts/set_bundle_build_number.sh:45-48`
- Nothing else reads `Build.xcconfig` during a run; `bump-version.sh` and `.github/workflows/bump-version.yml` are separate paths

A file that is never written cannot be left dirty, so this removes the failure mode rather than adding more trap coverage for cases that fundamentally can't be trapped.

## Kept

The self-heal for a stray `.release-bak` stays, so backups written by older versions of the script still get cleaned up (including any on machines right now) instead of lingering forever once nothing else writes them. It remains `--dry-run` gated.

## Verification

- `bash -n Scripts/release.sh` — syntax clean
- No dangling references to the removed `inject_build_number` / `restore_build_number` / `_injected`
- Confirmed the archive command array still carries both `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION`, so IPA build numbers are unchanged

**Worth a sanity check on the next real run** that the uploaded build number still matches `git rev-list --count HEAD` — that's the one behaviour this could plausibly affect, and I can't exercise the signing path from here.

---

> [!NOTE]
> **Medium Risk**
> CloudKit gating uses low-level Mach-O entitlement parsing on simulator and changes sync fallback behavior; release path removes a redundant xcconfig write but build numbers still depend on CLI overrides matching prior behavior.
> 
> **Overview**
> **CloudKit / iCloud:** Simulator entitlement detection no longer treats `NSUbiquitousContainers` in Info.plist as proof of CloudKit access. On simulator it now parses `com.apple.developer.icloud-container-identifiers` from the linker’s `__TEXT,__entitlements` Mach-O section (via MachO/`getsectiondata`), so unsigned or CI builds without entitlements are not treated as entitled. `SyncProviderFactory` and tvOS `CloudSyncStatusView` stop calling `CKContainer(identifier:)` directly; they use `iCloudConstants.container` and **fall back** to iCloud Documents sync (or skip the availability check) when the container is nil, avoiding the runtime trap on unentitled binaries.
> 
> **Release script:** `Scripts/release.sh` no longer injects `CURRENT_PROJECT_VERSION` into `Build.xcconfig` or restores it via EXIT traps. Build numbers still come from the existing `xcodebuild` `CURRENT_PROJECT_VERSION="$BUILD_NUMBER"` overrides. Cleanup only removes temp ASC key dirs; interrupt handling no longer promises xcconfig restore. A one-time self-heal still restores `Build.xcconfig` from a leftover `.release-bak` from older script versions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3393a1caaf955c35a5c26a3c0aecdb367aed738. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Scripts/release.sh
+++ b/Scripts/release.sh
@@ -129,33 +129,21 @@ if ! $DRY_RUN && ! $NO_BUILD; then
     fi
 fi
 
-# ── Build number injection ────────────────────────────────────────────────────────
+# ── Build number ──────────────────────────────────────────────────────────────────
+# Build.xcconfig is deliberately NOT modified.
+#
+# Every xcodebuild invocation below already passes CURRENT_PROJECT_VERSION as a
+# command-line build-setting override, which outranks the xcconfig, and the
+# Info.plist resolves CFBundleVersion from that build setting (see
+# Scripts/set_bundle_build_number.sh). Editing the file was redundant.
+#
+# It was also unsafe: the edit had to be undone by an EXIT trap, and a trap
+# cannot run when the script is SIGKILLed or when a wedged xcodebuild swallows
+# the interrupt. That left a tracked file dirty — carrying an epoch build number
+# like 1785899074 instead of the commit count — plus an orphaned .release-bak,
+# until someone happened to run this script again. A file that is never written
+# cannot be left dirty, so the failure mode is gone rather than mitigated.
 XCCONFIG_BAK="$XCCONFIG.release-bak"
-inject_build_number() {
-    log "Setting build number $BUILD_NUMBER in $(basename "$XCCONFIG")"
-    # Back up the EXACT current file (not HEAD) so restore preserves any unrelated
-    # uncommitted edits — a blanket `git checkout` here would silently wipe them.
-    cp "$XCCONFIG" "$XCCONFIG_BAK"
-    _injected=true
-    # BSD sed (macOS): -i '' for in-place, matching Scripts/bump-version.sh.
-    sed -i '' "s/^CURRENT_PROJECT_VERSION[[:space:]]*=.*/CURRENT_PROJECT_VERSION = ${BUILD_NUMBER}/" "$XCCONFIG"
-}
-
-# Whether THIS run injected a build number (and therefore owns the backup).
-# Restoring must be conditional on this, not merely on the backup existing: a
-# leftover backup from a previously killed run is NOT ours, and a --dry-run that
-# restored it would both mutate the working tree and consume the one chance the
-# next real run had to self-heal.
-_injected=false
-
-restore_build_number() {
-    # Restore the pre-inject file so the build number isn't left in the working tree.
-    # `return 0` so this EXIT-trap never sets a nonzero exit code — the trap's
-    # status becomes the script's.
-    $_injected || return 0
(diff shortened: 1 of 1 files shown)


JoeMatt:
These review threads are stale — my mistake, and I've corrected it.

I branched this PR off `fix/icloud-sim-entitlement-crash` instead of `develop`, so it carried all six iCloud commits and reviewers were quite reasonably reviewing `iCloudConstants.swift`, `SyncProviderFactory.swift` and `CloudSyncStatusView.swift` in what is supposed to be a release-script change.

**This PR now contains only `Scripts/release.sh`** (rebased onto develop).

The findings themselves were valid and are all fixed on **#3646**, where those files belong:

| Finding | Status on #3646 |
|---|---|
| **CRITICAL** — `_cloudKitEntitlementFromProvisioningProfile()` appends a second `</plist>`, so the plist never parses and `?? true` assumes entitled | **Fixed.** The slice already ended past the closing tag. This meant the *device* entitlement check had never worked — sideloaded builds without the entitlement fell through to `true` and trapped. Excellent catch, and pre-existing rather than introduced here. |
| tvOS status stuck on "Checking sync status…" | **Fixed** — the guard now sets `isAvailable`/`syncStatus` to the terminal unavailable state instead of returning silently. Regression I introduced. |
| tvOS falls back to `iCloudContainerSyncer`, which the platform declares unsupported | **Fixed** — `iCloudSyncMode.iCloudDrive` is `#if !os(tvOS)` and its subtitle says "Not supported on tvOS", so returning it was wrong. `createSyncProvider` now returns `nil` there; it has no callers in the repo, so the signature change is contained. Regression I introduced. |
| View models still constructing `CKContainer` directly | Deliberately deferred — tracked in **#3648**. |

Resolving these here since the code they point at is no longer in this PR's diff. They remain live on #3646.

PR #3681, opened 2026-09-08
feat: add multi-selection to game library

bendevnull:
## Summary
- Add multi-selection mode to game-library grid, list, search, and shelf cells.
- Normalize selection IDs and keep selection scoped to visible/current games across filtering, refresh, and navigation.
- Add selected visuals, accessibility state, visible bulk select/deselect actions, and pure selection-state tests.

## Testing
- `git diff --check` — passed.
- Swift/Xcode tests — not run: this Windows host does not have `swift` or `xcodebuild` installed.

## Known limitations
- Full iOS build and test validation remains to be performed on macOS with Xcode.

### DIFF
--- a/PVUI/Sources/PVSwiftUI/Consoles/ConsoleGamesView+MultiSelect.swift
+++ b/PVUI/Sources/PVSwiftUI/Consoles/ConsoleGamesView+MultiSelect.swift
@@ -49,12 +49,15 @@ extension ConsoleGamesView {
         }
     }
 
+
     /// Wraps a game cell with a selection indicator overlay when multi-select is active.
     /// Selection checkmark is placed top-leading to avoid conflicting with the
     /// cloud sync indicator badge at top-trailing.
     @ViewBuilder
     func multiSelectOverlay(md5: String, @ViewBuilder content: () -> some View) -> some View {
-        let isSelected = gamesViewModel.selectedGameMD5s.contains(md5)
+        let normalizedID = md5.trimmingCharacters(in: .whitespacesAndNewlines).uppercased()
+        let isSelected = gamesViewModel.selectedGameMD5s.contains(normalizedID)
+        let selectionLabel = isSelected ? "Selected" : "Not selected"
         ZStack(alignment: .topLeading) {
             content()
                 .overlay {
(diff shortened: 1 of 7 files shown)


JoeMatt:
Thanks for this — multi-select is a frequently requested feature and the overall shape is good. A few things I'd want fixed before merge; I verified each against the branch.

### 1. The toolbar is never torn down when the view goes away

`MultiSelectToolbarState.shared.deactivate()` is only reached from the `isMultiSelectMode → false` branch of `onChange` (`ConsoleGamesView+MultiSelect.swift:159`), and there's no `onDisappear` teardown.

`RetroMultiSelectToolbar()` is rendered once at the `RetroMainView` root, above the tab bar, so it outlives any one console list. Enter multi-select, then switch console tabs or push a detail screen without tapping Done, and the floating toolbar keeps rendering over an unrelated screen with `onDelete` / `onMoveToSystem` still bound to the view model you left. The user then taps Delete and either nothing happens or it acts on the previous console's selection.

`ConsoleGamesView` already has an `onDisappear` for gamepad teardown (`ConsoleGamesView.swift:996`) — deactivating there would match.

### 2. `performBatchMove` moves the file before the DB write, and leaves related files behind

```swift
try FileManager.default.moveItem(at: sourceURL, to: destinationURL)

try realm.write {
    let thawedGame = game.thaw()
    thawedGame?.system = system
    ...
}
```

Two problems:

- **No rollback.** If `realm.write` throws after the move succeeded, the file is at the new path while the record still points at the old one. The game is unlaunchable and is reported as "failed", so the user has no idea it's half-moved.
- **Multi-disc games break.** Only `sourceURL` moves. `game.relatedFiles` (the `.cue`/`.bin` and multi-disc companions — the same thing this PR's new `GameCellModel.discCount` counts) stays in the old system's directory. `relatedFiles` isn't referenced anywhere in the file.

Moving `relatedFiles` alongside the primary file, and ordering it so a failure can't leave the two out of sync, would fix both.

### 3. Tests don't cover the new logic

The added `PVSwiftUITests` cases all exercise `SaveImportMatchingService` and `ROMTitleNormalizer`, which are pre-existing string helpers. Nothing covers `MultiSelectToolbarState` (`activate`/`deactivate`/`updateCount`) or `ConsoleGamesViewModel.toggleSelection(md5:)` / `enterMultiSelectMode()` / `exitMultiSelectMode()` — all plain `@MainActor` logic with no Realm or UI dependencies, so they're cheap to test. The batch operations are harder to cover and I wouldn't block on those.

### Nit

The `forPrimaryKey: md5 ?? md5.uppercased()` fallback lookup is repeated five times in `ConsoleGamesView+MultiSelect.swift` (lines 152, 210, 277, 314, 372). A small `private func game(forMD5:in:)` helper would keep those from drifting.

### What's already right

- `@ObservedObject` for the `MultiSelectToolbarState` singleton and `@StateObject` for the owned view model — both match the project rule.
- Haptics and `.romDragSource` are correctly `#if !os(tvOS)` / `#if os(iOS)` guarded, so this still compiles for tvOS.
- Selection keyed by MD5 `Set<String>` rather than retained Realm objects, with a live re-fetch before each operation — that neatly sidesteps thread-confinement and invalidated-object problems.
- No raw `com.provenance.*` identifier comparisons.

Items 1 and 2 are the blockers; 3 is per the repo's PR requirements. Happy to re-review once they're addressed.

PR #3683, opened 2026-09-12
fix(webserver): stop serving retroarch.cfg and scrub it on logout

kah-ja:
### What does this PR do

- Added a denylist check to both classes' existing `_checkFileExtension:`: a `.cfg` file is rejected unconditionally, independent of `allowedFileExtensions`. This covers listing, download, upload, move, delete and rename through either server.
- Added `RetroArchConfigManager.clearPersistedCredentials()`, reusing the existing `syncToRetroArch()` write path with an empty username and password, called from `logout()`. It rewrites only those two lines.

Not changed: `allowedFileExtensions` stays `nil`, since a ROM-extension allowlist risks blocking saves and screenshots the uploader also serves. No authentication was added. `PVModernWebServer.swift`'s `/files` (`:239`) and `/download` (`:298`) serve the same root with no extension filter either, but that server sits behind the `useModernWebServer` flag, which ships off; it would need the same denylist if that default changes.

Trade-off: blocking `.cfg` also stops backing up `retroarch.cfg` itself over WebDAV or the uploader. The in-app config editor remains, though it does not round-trip the whole file.

`PVWebServer.m` and both server classes

- Both `_checkFileExtension:` implementations and the missing scrub call were unchanged at `develop` HEAD (`93f49b40`); `git log --since=2026-07-01` shows no prior fix attempt.
- `swiftlint` on the changed Swift files: no new violation; the one warning in `RetroAchievementsClient.swift` is present on unpatched `develop` too.
- `PVCheevosTests` does not exercise `syncToRetroArch` or `logout`, so no existing test changes behaviour.
- Not executed: a Swift build or an Xcode run of the Objective-C change. Neither toolchain runs on this Linux host.

- `PVWebServer.m:186` sets `allowedFileExtensions = nil`; `GCDWebUploader.m:222` and `GCDWebDAVServer.m:166` treat `nil` as "allow every extension", so both servers serve any file under Documents, including `RetroArch/config/retroarch.cfg`.
- `RetroArchConfigManager.swift:149-161` writes the RetroAchievements password into that same `retroarch.cfg`, in cleartext.
- `RetroAchievementsClient.swift:203-206` `logout()` clears the Keychain store but never touches the config file, so the cleartext password stays on disk after sign-out.

Neither LAN server has built-in auth. `Authentication.swift:6,97` shows the project already moved this same password out of plaintext UserDefaults into Keychain for its own store. The RetroArch config copy sat outside that fix, and outside `_checkFileExtension:`, the gate both server classes already carry in matching implementations.

GHSA-hjgm-g737-vvxc ([found during penetration test](https://www.turingpoint.de/en/security-assessments/pentests/) by [turingpoint](https://www.turingpoint.de), reported privately, unpublished at time of writing)

n/a, no UI change.

### Questions

None.


JoeMatt (approved):
Verified the underlying issue is real. `retroarch.cfg` lives at `Documents/RetroArch/config/retroarch.cfg`, and both `GCDWebUploader` and `GCDWebDAVServer` are initialised with the Documents folder as their root (`PVWebServer.m:125`/`:129`), so the plaintext `cheevos_password` really was reachable over the LAN with no authentication.

I also confirmed the extension filter is not merely cosmetic: `_checkFileExtension:` gates the actual download handler in `GCDWebUploader.m:302`, which returns 403, not just the directory listing. And the PROPFIND call site at `GCDWebDAVServer.m:445` is a listing builder, not a COPY/MOVE recursion, so folder moves will not silently drop files.

Approving, because partial mitigation of a cleartext credential beats none. Three things worth a follow-up rather than blocking this:

**1. COPY/MOVE validates only the destination, so the block is bypassable.** `performCOPY:isMove:` checks `dstAbsolutePath` and never looks at the source:

```
MOVE /RetroArch/config/retroarch.cfg  ->  /pwned.txt
GET  /pwned.txt
```

That defeats the filter entirely. Worth validating the source name too. Note this same gap already lets hidden files be renamed out of `_allowHiddenItems`.

**2. The modern server is unpatched.** `PVModernWebServer` (Hummingbird, opt-in via `Defaults[.useModernWebServer]`) also browses the Documents root and has no extension filtering at all, so this fix does not cover users who enabled that toggle.

**3. Consider narrowing `cfg` to the filename `retroarch.cfg`.** Blocking the whole extension also blocks listing, downloading and re-uploading per-core override `.cfg` files, which is a legitimate WebDAV workflow. The credential exposure is specific to one filename.

On the Swift side: `credentialsManager` is `RetroCredentialsManager.shared`, the same store `syncToRetroArch()` reads, so by the time the queued block runs after `clearAll()` the overrides are strictly redundant. Harmless and arguably more explicit, so no change requested.

The failing `SPM Test (PVCheevos)` check is not caused by this PR. `PVCheevosTests.swift` has not compiled since the `RetroArchConfigManager` suite was added, because the swift-testing `@Suite` macro rejects an explicit `@available` attribute. This PR just happened to be the first to touch PVCheevos and trigger the per-module test job. Fixed separately on develop in 1e889afa25; all 30 tests now compile and pass.

JoeMatt:
Landed on `develop` as 266b6fb046 via `git cherry-pick -x`, with you retained as the commit author. Closing this PR as applied, not rejected.

